PR #31, opened 2016-04-07
moved strings to translation file

### LINKED ISSUE #26: Localization
Play can be localized through the `Messages` API. Currently there's a mix of string constants defined in the `conf/messages.en-US` file and in plain HTML mostly cause lazy but Ore can be localized by creating new files with language code extensions (`conf/messages.fr`).

Most strings in the views should be moved to language files.

malte-laukoetter:
should fix #26 for now


windy1:
Thanks for this. Can you rebase your branch and squash your commits please?

malte-laukoetter:
i think i create a new one with only one commit because i'm somehow not able to get the smash and merge done... #38

PR #38, opened 2016-04-08
moved strings to translation file

### LINKED ISSUE #26: Localization

malte-laukoetter:
should fix #26 for now

also see #31

new pull request because i need to learn git a bit more ;)


windy1:
Merged in 6fd3b71ba9625aa33cc0a136063b6ca14e31b3b7

PR #58, opened 2016-04-14
Update .gitignore

Grinch:
Updated to better adhere to what is in place for a typical Play Framework and SBT .gitignore. Including the addition of /conf/application.conf.

Signed-off-by: Steven Downer grinch@outlook.com


windy1:
I don't actually think this will do what you think it'll do. I don't think a `.gitignore`d file is actually ignored until it is removed from the index.

I think 

`git update-index --assume-unchanged conf/application.conf` is what you're looking for. Unfortunately you will have to do this every time you clone the repo.

Grinch:
Adding this file to the .gitignore will stop tracking it on new commits from that point onwards. If you need to commit changes, just do `git add conf/application.conf -f` and it will override .gitignore rules.

**EDIT:** This allows me to modify it locally and not have to worry about maybe commit some details I'd rather not want to.

.gitignore
```diff
@@ -11,3 +11,4 @@ uploads
 /RUNNING_PID
 /.settings
 .DS_Store
+/conf/application.conf
```
kashike: ~~Missing newline~~

Grinch:
@windy1 We're not aiming to remove it from the repo. Just to stop tracking changes since it's a file that those working on this may be modifying on a per-user basis.

windy1:
@Grinch Yeah from what I've been reading just adding a file to the gitignore isn't enough to stop tracking it, it actually needs to be removed otherwise changes will still be staged. I could be wrong but do you have anything to back that up?

kashike:
The better solution would be to maintain a `conf/application.conf.dist` and to ignore `conf/application.conf`, if possible.

Grinch:
@windy1 You are correct regarding this. Apologies. What @kashike would be needed in addition to this.

I suggest `application.conf.template` instead of `application.conf.dist` though for the default name.

windy1:
That sounds like a good solution to me. Do you want to just add that to this PR @Grinch?

Grinch:
I will later this evening. We'll have to make note in instructions for setting up Ore locally to copy application.conf.template to application.conf and specify the details needed.

windy1:
Yeah we could also just make a setup script that does it automatically that could include any other steps necessary now or in the future.

jamierocks:
https://www.gitignore.io/api/osx%2Clinux%2Cwindows%2Cscala%2Cplayframework%2Cintellij%2Ceclipse might be a better gitignore to use ;)

Grinch:
@jamierocks Didn't know gitignore.io was a thing. Thanks! Replaced it with that one plus our additions in its own spot at the bottom. :)

@windy1 Done.

jamierocks:
Yeah I tend to use gitignore.io and https://github.com/Danimoth/gitattributes for all my repos ;)

PR #103, opened 2016-05-04
Fix infinite redirects during failed permission checks.

JBYoshi:
Steps to reproduce:
- Log in to an account that does not have admin access.
- Go to `/admin/queue` or any other `/admin/*` address.

In Firefox, it says "The page isn't redirecting properly: Firefox has detected that the server is redirecting the request for this address in a way that will never complete."

Sorry if I'm not using proper style; this is the first Scala project I've worked on.

### DIFF
--- a/app/controllers/Actions.scala
+++ b/app/controllers/Actions.scala
@@ -17,10 +17,9 @@ import scala.concurrent.Future
 trait Actions {
 
   def onUnauthorized(request: RequestHeader) = {
-    request.flash.get("noRedirect") match {
-      case None => Redirect(routes.Users.logIn(None, None, Some(request.path)))
-      case Some(t) => Redirect(routes.Application.showHome(None, None, None))
-    }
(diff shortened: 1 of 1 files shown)


app/controllers/Actions.scala
```diff
@@ -16,12 +16,10 @@ import scala.concurrent.Future
   */
 trait Actions {
 
-  def onUnauthorized(request: RequestHeader) = {
-    request.flash.get("noRedirect") match {
-      case None => Redirect(routes.Users.logIn(None, None, Some(request.path)))
-      case Some(t) => Redirect(routes.Application.showHome(None, None, None))
-    }
-  }
+  def onUnauthorized(request: RequestHeader) =
```
windy1: Multiline methods should use braces.

JBYoshi: Fixed.

PR #104, opened 2016-05-05
Fix download link for version downloads.

JBYoshi:
When you try to download a specific version of a plugin whose URL does not match the raw name (e.g. if the name contains spaces), the link doesn't work (example: https://ore-staging.spongepowered.org/djxy/Permission-Manager/versions/1.0). These links use the slug for the file name, while all the rest use the project name instead. This PR fixes that.


windy1:
Oh whoops, yeah this should have been using the name to begin with. Nice catch.

JBYoshi:
@windy1 Thanks :smiley:

PR #128, opened 2016-06-03
Use round avatars for consistency with forums.

JBYoshi:


Grinch:
We opted to go with square to match that of GitHub.

JBYoshi:
Ah, OK. Does that mean the forums are OK as they are, or should they be changed instead?

windy1:
There's no reason that Ore and the forums need to be 100% consistent with each other. I am also happy with the current look as I think it fits the rest of the site better.

PR #154, opened 2016-10-27
Use Sponge dropdown style from SpongeHome

stephan-gh:
I replaced the current left navigation with the complete code from SpongeHome (I've just copied and pasted the needed SCSS files), that should make it easier to update it in the future.

With these changes, Ore will also use the new (uppercase) style for the Sponge dropdown:

screen shot 2016-10-27 at 13 36 22

### DIFF
--- a/app/assets/stylesheets/_nav.scss
+++ b/app/assets/stylesheets/_nav.scss
@@ -1,55 +1,23 @@
 @import 'utils';
 
 // left nav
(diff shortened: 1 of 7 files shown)


windy1:
This has been reverted because you removed some CSS that was used elsewhere in the right side of the header (when you're logged in). Can you please fix it and open a new PR?

PR #155, opened 2016-10-29
Fix problems with new top bar

stephan-gh:
Fix for https://github.com/SpongePowered/Ore/pull/154#issuecomment-256734478

I fixed all problems I noticed: the background color when hovering, and the missing fixed width of the dropdown menu of the right navigation. Please let me know if there is anything else that needs to be fixed.

### DIFF
--- a/app/assets/stylesheets/_nav.scss
+++ b/app/assets/stylesheets/_nav.scss
@@ -23,6 +23,35 @@
 
 // right nav
 
+.navbar-inverse .navbar-nav > .open > a {
+   &, &:hover, &:focus, &:active {
+     background: transparent;
+  }
+}
+
+.navbar-nav > li > .main-dropdown, .user-dropdown {
+  @include no-box-shadow();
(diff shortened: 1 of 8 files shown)


windy1:
The drop downs are now a bit misaligned for me and can we just get rid of that darkened background on the pofile pic while we're at it (chrome):

image

See how it overlaps the header a bit? Also, drop down no longer opens when you hover but requires a click now, but tbh I think I prefer this anyways.

I also noticed you need to click the caret to open the "plus" drop down. You should be able to just hit the icon as well.

e: Also seems to break search

stephan-gh:
@windy1 Pushed some fixes, let me know if that fixes the issue for you.

windy1:
@Minecrell Looks good! Just one more thing though, the hover backgrounds on the right-side drop downs seems to have disappeared.

stephan-gh:
@windy1 Should be fixed now.

PR #196, opened 2017-01-09
Add a useful class to sponsor blocks

lol768:
Though there are no CSS rules defined in Ore's code for this class, it is used to convey the semantic meaning of the element's contents to popular browser extensions which make use of an external database of DOM and URL based rules.


windy1:
There are existing rules for both `sponsor` and `sponsor-featured` 

https://github.com/SpongePowered/Ore/blob/0cddaff7a8b9cf526d1ff475667d8360376c5140/app/assets/stylesheets/main.scss#L47-L62

E: Oh I understand what you mean. Can you please give an example on how adding this class is a benefit?

PR #200, opened 2017-01-14
Tweak SpongeAuth endpoint URLs

lukegb:
The SpongeAuth endpoint URLs changed a little bit with lukegb/SpongeAuth, so they need to be updated here.


lukegb:
Woo, PR #200!

PR #202, opened 2017-01-16
Allow the loading while being offline and add to Homescreen support

malte-laukoetter:
I have added a serviceworker to ore that will allow the page to be viewed while someone has no internet connection, if the page was already loaded previously and the page cached.

Currently most get requests are getting cached and post request are not, also the api is not cached. If someone tries to load a not cached page while being offline the [offline error page](https://github.com/Lergin/Ore/blob/feature/offline/app/views/errors/error.scala.html) is shown. Also a alert is shown on all pages if the user is offline so it is clear that the data may be obsolete.

The exact configuration of this can be viewed and changed in the sw.js file [https://github.com/Lergin/Ore/blob/feature/offline/public/sw.js](url)

The serviceworker is using the [serviceworker toolbox from google](https://github.com/GoogleChrome/sw-toolbox) and i don't really know how to place the files of it into the project so i just copied them into the public/sw-toolbox directory...

Also this adds a [WebApp manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest) with some configuration settings for the browser if the page is opened from the homescreen. Currently i set the settings so that the site opens without the browser ui and uses the ore logo as an icon.  This also colors the address bar of the browser in the color of the ore navigation bar. (https://github.com/Lergin/Ore/blob/feature/offline/public/manifest/manifest.json)

Through these changes users should also be prompted with a "Add to Homescreen" banner if they have the site open for some time on a mobile device (https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android).


public/javascripts/main.js
```diff
@@ -163,3 +163,15 @@ $(function() {
...
+ * ==================================================
+ * =                 Service Worker                 =
+ * ==================================================
+ */
+
+if ('serviceWorker' in navigator) {
+    window.addEventListener('load', function() {
+        navigator.serviceWorker.register('/sw.js', {scope: '/'});
+    });
+}
```
windy1: Missing newline.

public/sw.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function networkFirstIndex(request, values, options) {
+  return toolbox.networkFirst(request, values, options).catch((err) => {
+    return toolbox.cacheOnly(new Request('/'));
+  });
+}
+
+// pre cache all urls that are needed for the main and offline page
+var preCache = [
```
windy1, line 22: My only problem here and below is maintainability, is there a way we can make this more coherent so we don't need to update this script every time we add a new route? You can make a Play template that can gets the route `documentation` from the `Router` and maybe just keep a list of excludes here. It might seem overcomplicated now but it would be a great feature to have for future-proofing.

windy1 (changes requested):
Great idea!

public/sw.js
malte-laukoetter: The only really hard to maintain part of this is the preCache but i don't know how to get all the routes of the pages in there automatically and haven't really worked with scala or play... 

The rest of the routes shouldn't needed to be touched because all post and get requests are getting send to the right router as long as there is no problem with caching the get requests (i.e. the request only gets data and doesn't change anything on the server), the reason for the list of special cases is that some urls that are GET request really should be POST requests (e.g. admin/seed) and some are downloads that are too large to be cached without problems. Also some of the other routes there shouldn't really be necessary so i deleted some of them in a commit.

windy1 (approved):

windy1:
Merged in 9189ec09b8b6ddc384b896c6ffb3372a36bd73f6

PR #226, opened 2017-02-28
Fix download button in IE and Edge

Flibio:
Fixes the download button issue in Internet Explore and Edge.


windy1:
Thank you for the contribution but the feature/downloads branch changes the download routes back to GET so this is actually not needed.

Flibio:
Okay!

PR #238, opened 2017-03-13
Use webjars for jquery, FA, filesize and sw-toolbox

stephan-gh:
Currently, jquery, font awesome and some other libraries are directly bundled in the repository. That makes them more difficult to update. The [WebJars](http://www.webjars.org/) project packages these libraries in JAR files and hosts them on Maven Central.

Using Play Framework, all that is necessary to use webjars is to define a dependency on it. It can then be referenced from the assets route in `lib/<name>/path/to/file.js`. We can use Webjars for jquery, font-awesome, sw-toolbox and filesize. Bootstrap and highlight.js cannot be bundled using webjars currently because they are custom packages.


windy1:
> +23
> -2,447
> 🙌

Issue #241, opened 2017-03-14
Dumb security audit


ImMorpheus:
- [x] We should add all the fancy secure headers; 

- X-Frame-Options

- X-XSS-Protection

- X-Content-Type-Options

- [x] Currently Ore set three cookies (I don't think I missed something): PLAY_SESSION, _stat and _warning. All of them are set without the secure flag )

- [ ] No Cache control header is used (we should use no-cache, no-store, etc.) (I'm talking about the page, it's not relevant if we cache certain images like avatar or such

- [x] Like the home (soon), ore should use a CSP: Something like this: 
Content-Security-Policy: default-src 'none'; script-src 'self' 'unsafe-inline'; style-src 'self'; img-src 'self' https://auth.spongepowered.org/ https://forums-cdn.spongepowered.org/ https://www.gravatar.com/ https://www.spongepowered.org/; font-src https://fonts.googleapis.com/; connect-src 'self'; form-action 'self'; upgrade-insecure-requests; block-all-mixed-content; 
NB: do not use this, this is just a draft. It should be fixed, for example it doesn't allow img to be loaded in the plugin page (people use imgur or such to upload their images)

Ok, I didn't have much time to test extensively and most problems were found with auth.spongepowered.org so maybe I should open an issue there. 
Also, I got an sql error while playing with &token= in the download page, is it checked for invalid character ? (EDIT: nevermind, added down below. Scroll the comments)
I also found some minor things, not really issues, like [this](https://ore.spongeproject.net/?page=2147483648)
I'm not sure I have finished, I will update this tomorrow

lol768:
>  Also, I got an sql error while playing with &token= in the download page, is it checked for invalid character ?

Which page/proof of concept would be useful. Can't reproduce on Ore, anyway.

Most of the headers have already been (privately) suggested. Not sure why caching is really a security problem. Cookies could do with sorting if there's no HSTS policy in place.

lukegb:
https://hstspreload.org/?domain=spongepowered.org
https://hg.mozilla.org/mozilla-central/raw-file/tip/security/manager/ssl/nsSTSPreloadList.inc

lukegb:
SpongeAuth is currently at lukegb/SpongeAuth, please feel free to report issues there 👍

ImMorpheus:
[...]

PR #243, opened 2017-03-15
Set some HTTP security headers

stephan-gh:
Configures Play's [SecurityHeadersFilter](https://www.playframework.com/documentation/2.5.x/SecurityHeaders) to set some common HTTP headers:

```
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Permitted-Cross-Domain-Policies: master-only
```

CSP (`Content-Security-Policy`) is useful to prevent all kinds of XSS attacks, however it is a bit difficult to configure it for Ore: Ore uses a lot of inline scripts and inline styles and replacing all of them is out of scope for this PR (some of them can't be replaced easily). Consequently, it isn't too strict:

- Scripts are either loaded from JavaScript files hosted on Ore, or specified inside inline scripts with an explicit nonce (a random Base64 encoded string). A new nonce is generated for each request, so injecting a simple `<script>` tag isn't possible anymore.
- For CSS styles, `unsafe-inline` is allowed. Ore uses a lot of inline styles and these must be refactored into the separate CSS files before it can be made more strict.
- Images can be loaded from everywhere, but only using `https://`. I'm not entirely sure if images from `http://` are allowed currently. (All in all, mixed content is pretty bad though)

I don't have the discourse integration locally so I wasn't able to test the rules for that.

Fixes a part of #241

### DIFF
--- a/app/Filters.scala
+++ b/app/Filters.scala
@@ -2,5 +2,8 @@ import javax.inject.Inject
 
 import play.api.http.DefaultHttpFilters
 import play.filters.csrf.CSRFFilter
+import play.filters.headers.SecurityHeadersFilter
+import security.NonceFilter
 
-class Filters @Inject()(csrfFilters: CSRFFilter) extends DefaultHttpFilters(csrfFilters)
+class Filters @Inject()(csrfFilters: CSRFFilter, securityHeadersFilter: SecurityHeadersFilter, nonceFilter: NonceFilter)
+  extends DefaultHttpFilters(csrfFilters, nonceFilter, securityHeadersFilter)
--- a/app/security/NonceFilter.scala
+++ b/app/security/NonceFilter.scala
@@ -1,0 +1,33 @@
+package security
+
+import java.security.SecureRandom
+import java.util.Base64
+import javax.inject.Inject
+
+import akka.stream.Materializer
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.mvc.{Filter, RequestHeader, Result}
+
+import scala.concurrent.Future
+
+object NonceFilter {
+  def nonce(implicit request: RequestHeader): String = request.tags("nonce")
+}
+
+class NonceFilter @Inject() (implicit val mat: Materializer) extends Filter {
+  private val random = new SecureRandom()
+
+  override def apply(next: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    val nonce = generateNonce
+    next(request.withTag("nonce", nonce)).map { result =>
+      result.withHeaders("Content-Security-Policy" -> result.header.headers("Content-Security-Policy")
+        .replace("%NONCE-SOURCE%", s"nonce-$nonce"))
+    }
+  }
+
+  private def generateNonce: String = {
+    val bytes = new Array[Byte](16) // 128 bit, see https://w3c.github.io/webappsec-csp/#security-nonces
+    random.nextBytes(bytes)
+    Base64.getEncoder.encodeToString(bytes)
+  }
+}
--- a/app/views/bootstrap/layout.scala.html
+++ b/app/views/bootstrap/layout.scala.html
@@ -6,6 +6,7 @@ this.
 @import db.impl.access.UserBase
 @import ore.OreConfig
 @import play.twirl.api.Html
+@import security.NonceFilter._
 @(title: String, scripts: Boolean = true, header: Boolean = true, user: Boolean = true,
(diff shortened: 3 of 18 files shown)


lol768:
LGTM broadly.

Out of scope for this, but image loading in user content should work via a TLS image proxy (with some sort of signature to prevent misuse) - else UX will be poor here. Consideration: leaking of backend IPs.

CSP w/ nonce is a good win for security - but are there any `<script>` elements with dynamic (user) content inside? I'm sure these are properly handled by JSON encoding everything, but this would be my first attempt to bypass this.

ImMorpheus:
I've some doubts about this.

There's no need to specify both 'self' and https, if you want to allow images to be loaded from everywhere, then the domain is included. 
Fonts are only loaded from https://fonts.gstatic.com/ right ? (not entirely sure)

> I'm not entirely sure if images from http:// are allowed currently

No, you specified block-all-mixed-content, thus resources can't be loaded from http if the page is served over https (which is always done because of hsts and hsts preloading).

What about something like this ? 

Content-Security-Policy: default-src 'none'; script-src 'self' %NONCE-SOURCE% https://www.google-analytics.com/analytics.js; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; img-src *; font-src https://fonts.gstatic.com/; connect-src 'self'; frame-ancestors 'none'; form-action 'self'; upgrade-insecure-requests; block-all-mixed-content; 

NB:
- default-src has been set to none because that's what I usually do, as child-src, connect-src, font-src, img-src, media-src, object-src, script-src and style-src fall back to default-src if they are not specified. So by using default-src 'self' I'd implicitly set to 'self' all of them. If we want to change this to self, it's better to explicitly set what we don't use to none.
- img-src is set to all and nothing else, because the https part is taken care by upgrade-insecure-requests; block-all-mixed-content;, avoiding situations like [this](https://ore.spongepowered.org/Flibio/EconomyLite)
- form-action is set to self, but I'm not 100% sure there aren't any form submissions to auth.spongepowered.org. 
- frame-ancestors just replace X-Frame-Options header, conventions say it should be preferred
- connect-src it's because of the XMLHttpRequest for the spongemark to ore.spongepowered.org (so self)

lol768:
> frame-ancestors just replace X-Frame-Options header, conventions say it should be preferred

You're not wrong, but browsers which don't support CSPv2 will not be able to understand this directive. Sending both is probably preferable. I can't recall which browsers in particular don't understand the latest spec, but IE is probably there somewhere.

ImMorpheus:
Yeah, I was vague. The headers are ok, my point was that frame-ancestors should be sent even if we specified the X-Frame-Options, so that newer browser can use it as it "obsoletes" the header. 
Anyone know about form submits to other domains ? That's what is bothering me the most

stephan-gh:
> There's no need to specify both 'self' and https, if you want to allow images to be loaded from everywhere, then the domain is included.

There is, because Ore is not always served from `https://`, for example when running a local test instance.

> No, you specified block-all-mixed-content, thus resources can't be loaded from http if the page is served over https (which is always done because of hsts and hsts preloading).

The "currently" was referring to the state before the PR was merged. I'm well aware that insecure images are no longer loaded using the new rule.

> default-src has been set to none because that's what I usually do, as child-src, connect-src, font-src, img-src, media-src, object-src, script-src and style-src fall back to default-src if they are not specified. So by using default-src 'self' I'd implicitly set to 'self' all of them. If we want to change this to self, it's better to explicitly set what we don't use to none.

It's not like we have a really permissive `default-src`. I don't see a real benefit in setting it to `none` except making the whole system more complicated. (We'd need to set `manifest-src`, `worker-src`, `form-action` and `connect-src`. The only thing that isn't used **currently** is `object-src` and `media-src` and I doubt they will cause any harm.)

> img-src is set to all and nothing else, because the https part is taken care by upgrade-insecure-requests; block-all-mixed-content;, avoiding situations like this

Why do you allow everything for `img-src`? There is no need to accept anything except `self` and `https:` if you have `upgrade-insecure-requests` because there should be never a resource that is loaded using an insecure connection (and that would be automatically blocked by `block-all-mixed-content`).

The only reason why I haven't enabled `upgrade-insecure-requests` yet is that it breaks the local test server because it's running on `http://` by default.

> frame-ancestors just replace X-Frame-Options header, conventions say it should be preferred

As @lol768 mentioned, both should be sent - I just forgot to configure it.

ImMorpheus:
I see, I assumed those settings were for a production server. If they are used for testing and you need http:// then my doubts are cleared and those rules are justified, solving all the points.
My bad for the 'currently'.

PR #252, opened 2017-03-31
Add a 404 error handler page

simon816:
Looks like this:

404 page

There is a minor issue when the page is shown after falling back to the [`ErrorHandler`](https://github.com/simon816/Ore/blob/bec33404f98228dae6c7c2b1c6a7badcd9e33ff6/app/ErrorHandler.scala#L41-L47) , the `Request[_]` implicit variable for [`layout.scala.html`](https://github.com/SpongePowered/Ore/blob/master/app/views/bootstrap/layout.scala.html#L11) is null because we are only given a `RequestHeader` variable. This causes the [CSRF token](https://github.com/SpongePowered/Ore/blob/master/app/views/bootstrap/layout.scala.html#L67-L69) and the [nonce](https://github.com/SpongePowered/Ore/blob/master/app/security/NonceFilter.scala#L15-L19) values to not exist on the page.
The fix would be to make all `Request[_]` parameters into `RequestHeader`, but I didn't want to be too disruptive.

### DIFF
--- a/app/controllers/BaseController.scala
+++ b/app/controllers/BaseController.scala
@@ -31,6 +31,8 @@ abstract class BaseController(implicit val env: OreEnv,
 
   override val signOns: ModelAccess[SignOn] = this.service.access[SignOn](classOf[SignOn])
 
+  override def notFound()(implicit request: Request[_]) = NotFound(views.html.errors.notFound())
+
   /**
     * Executes the given function with the specified result or returns a
     * NotFound if not found.
@@ -41,8 +43,8 @@ abstract class BaseController(implicit val env: OreEnv,
     * @param request  Incoming request
     * @return         NotFound or function result
     */
-  def withProject(author: String, slug: String)(fn: Project => Result)(implicit request: RequestHeader): Result
-  = this.projects.withSlug(author, slug).map(fn).getOrElse(NotFound)
+  def withProject(author: String, slug: String)(fn: Project => Result)(implicit request: Request[_]): Result
+  = this.projects.withSlug(author, slug).map(fn).getOrElse(notFound)
 
   /**
     * Executes the given function with the specified result or returns a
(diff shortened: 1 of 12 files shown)


app/controllers/BaseController.scala
```diff
@@ -31,6 +31,10 @@ abstract class BaseController(implicit val env: OreEnv,
 
   override val signOns: ModelAccess[SignOn] = this.service.access[SignOn](classOf[SignOn])
 
+  override def notFound()(implicit request: Request[_]) = {
```
windy1: Do you actually need this AND the ErrorHandler override?

app/controllers/sugar/Actions.scala
```diff
@@ -298,8 +305,14 @@ trait Actions extends Calls with ActionHelpers {
     def refine[A](request: AuthRequest[A]) = Future.successful {
       Actions.this.organizations.withName(organization)
         .map(new AuthedOrganizationRequest[A](_, request))
-        .toRight(NotFound)
+        .toRight(notFound0(request))
     }
   }
 
+  // Convert explicit parameter to implicit
+  private def notFound0(request: Request[_]) = {
```
windy1: This is not necessary, you can actually pass implicit parameters explicitly.

```scala
notFound()(request);
```

will work fine.

app/views/errors/notFound.scala.html
```diff
@@ -0,0 +1,15 @@
+@import db.ModelService
+@import db.impl.access.UserBase
+@import ore.{OreConfig, OreEnv}
+@()(implicit messages: Messages, session: Session, requestHeader: RequestHeader, request: Request[_] = null,
+        service: ModelService, config: OreConfig, users: UserBase, env: OreEnv)
+
+@bootstrap.layout("404 Not Found", user = false) {
```
windy1: Please add all constant strings to the `messages` file and access through `Messages`.

windy1 (changes requested):

app/controllers/BaseController.scala
simon816: It's probably possible to delegate this to the ErrorHandler if it was injected but this was easiest. I could change it if you would like

windy1 (approved):

PR #274, opened 2017-07-14
Add warning for external links on pages. Closes #218

### LINKED ISSUE #218: Add interstitial page for external links?
Prior art:

* https://www.curseinc.com/linkout?remoteUrl=/
* https://minecraft.curseforge.com/linkout?remoteUrl=/
* tools.bukkit.org (dead, but ask lukegb)

Would prevent:

* Disguised links to off-site downloads
* Certain types of phishing links

Disadvantages:

* Might annoy some users (preference? logged-out users?)
* Bypassable by sticking to unverified download links on Ore (see https://ore.spongepowered.org/lol768/Maintest)

simon816:
Closes #218

image

Design based on the warning for un-reviewed plugin files.

Set the `app.trustedUrlHost` config option to the trusted host

The "get me out of here" link simply goes to the homepage, maybe it could go to the `Referer` header if exists or "javascript:history.go(-1)"?

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -84,6 +84,16 @@ final class Application @Inject()(data: DataHelper,
   }
 
   /**
+    * Show external link warning page.
+    *
+    * @return External link page
+    */
+  def linkOut(remoteUrl: String) = {
+    Action { implicit request =>
+      Ok(views.linkout(remoteUrl))
+    }
+  }
+  /**
     * Shows the moderation queue for unreviewed versions.
     *
     * @return View of unreviewed versions.
--- a/app/models/project/Page.scala
+++ b/app/models/project/Page.scala
@@ -166,7 +166,9 @@ object Page {
     }
   }
 
-  private val (markdownParser, htmlRenderer) = {
+  private var linkResolver: Option[LinkResolverFactory] = None
+
+  private lazy val (markdownParser, htmlRenderer) = {
     val options = new MutableDataSet()
       .set[java.lang.Boolean](HtmlRenderer.SUPPRESS_HTML, true)
 
@@ -187,13 +189,18 @@ object Page {
         TypographicExtension.create()
       ))
 
+
     (Parser.builder(options).build(), HtmlRenderer.builder(options)
-      // TODO: Get OreConfig using dependency injection or pass as parameter
-      .linkResolverFactory(new ExternalLinkResolver.Factory(play.api.Play.current.injector.instanceOf(classOf[OreConfig])))
+      .linkResolverFactory(linkResolver.get)
       .build())
   }
 
-  def Render(markdown: String): Html = Html(htmlRenderer.render(markdownParser.parse(markdown)))
+  def Render(markdown: String)(implicit config: OreConfig): Html = {
+    // htmlRenderer is lazy loaded so linkResolver will exist upon loading
+    if (linkResolver.isEmpty)
+      linkResolver = Some(new ExternalLinkResolver.Factory(config))
+    Html(htmlRenderer.render(markdownParser.parse(markdown)))
+  }
 
   /**
     * The name of each Project's homepage.
--- a/app/models/project/Page.scala
+++ b/app/models/project/Page.scala
@@ -146,18 +146,23 @@ object Page {
     private def wrapExternal(urlString: String) = {
       try {
         val uri = new URI(urlString)
-        if (uri.getScheme == null || "http" == uri.getScheme || "https" == uri.getScheme) {
-          val host = uri.getHost
-          val trustedUrlHost = this.config.app.getString("trustedUrlHost").get
-          if (host == null || host.endsWith(trustedUrlHost)) {
+        val host = uri.getHost
+        if (uri.getScheme != null && host == null) {
+          if (uri.getScheme == "mailto") {
             urlString
           } else {
             controllers.routes.Application.linkOut(urlString).toString()
           }
         } else {
-          urlString
+          val trustedUrlHosts = this.config.app.getStringSeq("trustedUrlHosts").get
+          val checkSubdomain = (trusted: String) => trusted(0) == '.' && (host.endsWith(trusted) || host == trusted.substring(1));
+          if (host == null || trustedUrlHosts.exists(trusted => trusted == host || checkSubdomain(trusted))) {
+            urlString
+          } else {
+            controllers.routes.Application.linkOut(urlString).toString()
+          }
         }
-      } catch { case _: URISyntaxException => urlString }
+      } catch { case _: URISyntaxException => controllers.routes.Application.linkOut(urlString).toString() }
     }
   }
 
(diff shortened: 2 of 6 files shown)


app/models/project/Page.scala
```diff
@@ -141,7 +182,10 @@ object Page {
         TypographicExtension.create()
       ))
 
-    (Parser.builder(options).build(), HtmlRenderer.builder(options).build())
+    (Parser.builder(options).build(), HtmlRenderer.builder(options)
+      // TODO: Get OreConfig using dependency injection or pass as parameter
+      .linkResolverFactory(new ExternalLinkResolver.Factory(play.api.Play.current.injector.instanceOf(classOf[OreConfig])))
```
simon816: Not sure how to get an instance of  `OreConfig` here. Maybe pass as implicit parameter to `Render(String)` below and build a new `htmlRenderer` with the config?

lol768:
So what stops me linking to `totallylegitspongepowered.org/phishing-page`

app/models/project/Page.scala
```diff
@@ -120,6 +123,44 @@ case class Page(override val id: Option[Int] = None,
...
+        link
+      } else {
+        link.withStatus(LinkStatus.VALID).withUrl(wrapExternal(link.getUrl))
+      }
+    }
+
+    private def wrapExternal(urlString: String) = {
+      try {
+        val uri = new URI(urlString)
+        if (uri.getScheme == null || "http" == uri.getScheme || "https" == uri.getScheme) {
```
lol768: Are external links to `ftp://` URLs considered okay? Why do we even need to check the scheme?

conf/application.conf.template
```diff
@@ -9,6 +9,7 @@ application {
     db.default-timeout  =   10
     uploadsDir = "uploads"
     uploadsDir = ${?UPLOADS_DIR}
+    trustedUrlHost      =   "spongepowered.org"
```
lol768: This would be better as a list IMO

app/models/project/Page.scala
```diff
@@ -120,6 +123,44 @@ case class Page(override val id: Option[Int] = None,
...
+          val trustedUrlHost = this.config.app.getString("trustedUrlHost").get
+          if (host == null || host.endsWith(trustedUrlHost)) {
+            urlString
+          } else {
+            controllers.routes.Application.linkOut(urlString).toString()
+          }
+        } else {
+          urlString
+        }
+      } catch { case _: URISyntaxException => urlString }
```
lol768: Don't default to untrusted user input for a failure condition.

Your current implementation is trivially bypassable simply by appending `?a""` to the end of the link URL.

lol768 (changes requested):

app/models/project/Page.scala
simon816: So you're saying that all links should be wrapped? even things like `mailto:` and `tel:`? with mailto it would allow all `@spongepowered.org` unless mail links were to be handled differently.

lol768: I think you could reasonably make exceptions for those two, but it should definitely be a whitelist (rather than a `http`/`https` blacklist).

simon816: Actually, I think I'll make all links go to the warning page, then the continue button will just link to whatever it is (including mailto)

lol768: Fair enough, I'm indifferent to both approaches. Always using the warning page is safer for sure

simon816:
Addressed the issued raised by @lol768

lol768:
LGTM now, @simon816

jamierocks:
> The "get me out of here" link simply goes to the homepage, maybe it could go to the Referer header if exists or "javascript:history.go(-1)"?

Eww, have the page load on a new tab, and have the "get me out of here" link close tab.

app/models/project/Page.scala
```diff
@@ -120,6 +123,49 @@ case class Page(override val id: Option[Int] = None,
 
 object Page {
 
+  private object ExternalLinkResolver {
```
windy1: If you make this a class (in a new file) you can use DI to get an instance of OreConfig and then use DI in a controller to get an instance of this class.

app/models/project/Page.scala
windy1: See above

windy1 (changes requested):

app/models/project/Page.scala
simon816: That would still mean rebuilding the `HtmlRenderer` for each invocation though. Don't think I can use DI in `private val (markdownParser, htmlRenderer) `

app/models/project/Page.scala
```diff
@@ -141,10 +189,18 @@ object Page {
...
+    (Parser.builder(options).build(), HtmlRenderer.builder(options)
+      .linkResolverFactory(linkResolver.get)
+      .build())
   }
 
-  def Render(markdown: String): Html = Html(htmlRenderer.render(markdownParser.parse(markdown)))
+  def Render(markdown: String)(implicit config: OreConfig): Html = {
+    // htmlRenderer is lazy loaded so linkResolver will exist upon loading
+    if (linkResolver.isEmpty)
+      linkResolver = Some(new ExternalLinkResolver.Factory(config))
```
simon816, line 201: I've removed the use of getting the OreConfig out of the global state. Still don't like this solution but it works

conf/application.conf.template
```diff
@@ -9,6 +9,7 @@ application {
     db.default-timeout  =   10
     uploadsDir = "uploads"
     uploadsDir = ${?UPLOADS_DIR}
+    trustedUrlHosts     =   [ ".spongepowered.org" ]
```
phase, line 12: Add a comment here saying that these need to start with a `.`.

app/controllers/Application.scala
```diff
@@ -83,6 +83,16 @@ final class Application @Inject()(data: DataHelper,
...
+  /**
+    * Show external link warning page.
+    *
+    * @return External link page
+    */
+  def linkOut(remoteUrl: String) = {
+    Action { implicit request =>
+      Ok(views.linkout(remoteUrl))
+    }
+  }
```
phase, line 95: Add a new line after this.

phase (changes requested):
This PR looks good. There are some little formatting things, but other then that this is good to merge.

app/views/linkout.scala.html
```diff
@@ -0,0 +1,34 @@
...
+            <div class="col-md-6 col-md-offset-3">
+                <div class="panel panel-default">
+                    <div class="panel-heading">
+	                    <h3 class="panel-title">
+	                        @messages("general.linkout.title")
+	                    </h3>
+                    </div>
+                    <div class="panel-body">
+                        <p>@messages("general.linkout.warning", remoteUrl)</p>
+                        <a href="/" class="pull-left unsafe-dl-back">
```
phase, line 20: This needs to go back to the previous page instead of going home.

jamierocks: I've said previously, but I think the best approach is to
> have the page load on a new tab, and have the "get me out of here" link close tab.

progwml6: i prefer phases approach on this one, with some of the isolation stuff coming to browsers I am not sure how reliable that will be long term

progwml6:
#325

Issue #272, opened 2017-07-13
Maximal amount of wiki pages


NeumimTo:
Is it possible to increase wiki pages limit? Currently we can have only 10.

NeumimTo:
https://github.com/SpongePowered/Ore/blob/ac577384f5648a8ec83b7baba8e6c10c5543c99f/app/views/projects/pages/view.scala.html#L62 

after a little digging i found out that this is just configuration.

phase:
This can be closed I believe.

PR #279, opened 2017-08-03
Bump default max number of wiki pages to 50

lol768:
#272 

Will likely require a conf change for prod.

### DIFF
--- a/conf/application.conf.template
+++ b/conf/application.conf.template
@@ -87,7 +87,7 @@ ore {
     
     projects {
         max-name-len = 25
-        max-pages = 10
+        max-pages = 50
         max-channels = 5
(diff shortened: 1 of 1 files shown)


simon816:
According to @NeumimTo on the forums (https://forums.spongepowered.org/t/wiki-pages/19838/8) the max page count is only verified client side, maybe you could add a check server side too.

NeumimTo:
Yep thats correct. No validation in server's bussines logic.

lol768:
Yeah, mbaxter mentioned that to me... Will add a server-side check

lol768:
#284

kashike (approved):

PR #283, opened 2017-08-05
Show Spongie as default when no avatar available (e.g. when testing)

lol768:
When not using SSO, the avatarUrl appears to be `None`. This is a simple default.


JBYoshi (approved):
I actually thought about doing this in the past; I just never looked hard enough to find a version of the logo without the text.

kashike (approved):

PR #284, opened 2017-08-05
WIP: Server-side page max checking

lol768:
Problems:

* Redirects transparently followed by browser, `go(xhr.getResponseHeader('Location'));` doesn't work because Play returns a redirect.

### DIFF
--- a/app/controllers/project/Pages.scala
+++ b/app/controllers/project/Pages.scala
@@ -6,7 +6,7 @@ import controllers.BaseController
 import controllers.sugar.Bakery
 import db.ModelService
 import form.OreForms
-import models.project.Page
+import models.project.{Page, Project}
 import ore.permission.EditPages
 import ore.{OreConfig, OreEnv, StatTracker}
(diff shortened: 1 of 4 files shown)


lol768:
Why is this done with AJAX anyway?

simon816:
Looks like it's done with AJAX because of the javascript logic before the request

lol768:
>Looks like it's done with AJAX because of the javascript logic before the request

Can you think of any reason why I can't implement this in the controller?

simon816:
You probably could do it in the controller and just have a form instead

vectrixdevelops:
@lol768 Any update on this?

lol768:
@connorhartley 

Unlikely to have time to finish this on top of my security consulting work, sorry.

If anyone wants to pick this up, I suggest you remove the unnecessary AJAX and do this with a POST properly. You can probably reuse the code I have in the controller, it'll be a case of making the frontend stuff form-based.

lol768:
RIP

PR #288, opened 2017-10-27
Add Filter to Compress HTML

phase:
HTML on all pages will be compressed. This uses the most up to date HTML compression library I found on Maven.

### DIFF
--- a/app/util/HtmlCompressionFilter.scala
+++ b/app/util/HtmlCompressionFilter.scala
@@ -1,0 +1,59 @@
+package util
+
+import javax.inject.Inject
+
+import akka.stream.Materializer
+import akka.stream.scaladsl.{Keep, Sink, Source}
+import akka.util.ByteString
+import com.googlecode.htmlcompressor.compressor.HtmlCompressor
+import play.api.http.{HeaderNames, HttpEntity, HttpProtocol}
+import play.api.mvc._
+
+import scala.concurrent.{ExecutionContext, Future}
+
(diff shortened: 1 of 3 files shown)


phase:
[Here](https://gist.github.com/phase/6176b8f399cb1650576e9f1ff95ad57d) is a gist that has a compressed version and an uncompressed version. Going from 18,830 bytes to 9,302 bytes.

/cc @kashike

app/util/HtmlCompressionFilter.scala
```diff
@@ -0,0 +1,59 @@
...
+      } else {
+        Future.successful(result)
+      }
+    })
+  }
+
+  private def isCompressible(result: Result): Boolean = {
+    val isChunked = result.header.headers.get(HeaderNames.TRANSFER_ENCODING).contains(HttpProtocol.CHUNKED)
+    val isGzipped = result.header.headers.get(HeaderNames.CONTENT_ENCODING).contains("gzip")
+    !(isChunked || isGzipped)
```
stephan-gh, line 50: This looks like you avoid compressing the HTML if the response is GZIP compressed. Considering like 100% of all browsers support GZIP, it sounds like the actual issue here is that Ore doesn't compress responses using GZIP.

Katrix: Would maybe a better solution be to just add the Gzip filter?
https://www.playframework.com/documentation/2.6.x/GzipEncoding

Katrix:
Will be done on ngnix level instead, or if not that, using the provided play Gzip filter

PR #289, opened 2017-10-27
Add Downloads Counter for the Versions Page (Fixes #268)

### LINKED ISSUE #268: Per release download count?
currently the overview of a project in the API has a total downloads field `downloads`. It would nice to have similar statistic for each version

phase:
Versions now show how many downloads they have. I wasn't able to test this super thoroughly locally, but digging through the code suggests that this should work completely fine.

### DIFF
--- a/app/views/projects/versions/list.scala.html
+++ b/app/views/projects/versions/list.scala.html
@@ -90,6 +90,7 @@ Versions page within Project overview.
                                                     <div class="pull-right">
                                                         <span class="date">@prettifyDate(version.createdAt.get)</span>
                                                         <span class="date">@version.humanFileSize</span>
+                                                        <span class="date">@version.downloadCount Downloads</span>
                                                         <a href="@versionRoutes.download(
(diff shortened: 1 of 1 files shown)


app/views/projects/versions/list.scala.html
```diff
@@ -90,6 +90,7 @@ <h4 class="panel-title pull-left">@messages("project.versions")</h2>
                                                     <div class="pull-right">
                                                         <span class="date">@prettifyDate(version.createdAt.get)</span>
                                                         <span class="date">@version.humanFileSize</span>
+                                                        <span class="date">@version.downloadCount Downloads</span>
```
kashike: This will display `1 Downloads` - should be `1 Download`.

kashike (changes requested):

app/views/projects/versions/list.scala.html
vectrixdevelops: I think this should be above the date (to the left of date) as it isn't as important as file size and date published to be next to the button to download the release in my opinion.

PR #292, opened 2017-10-31
WIP: Scrape README from GitHub if the source is provided

### LINKED ISSUE #87: On create project, optionally pull README.md content from the GitHub repo
When creating a new project, and I provide a GitHub repository URL, Ore should be able to pull the contents of my README.md file down from my GitHub repo.

Something like this would be a huge time-saver for a lot of developers. NPM for Node.js does something similar.

phase:
If a user provides a source repository on GitHub for the project, the
contents of the Home page will be set to the contents of the README.

Closes #87


app/util/GitHubUtil.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import java.io.FileNotFoundException
+
+import scala.io.Source
+
+object GitHubUtil {
+
+  private val identifier = "A-Za-z0-9-_"
+  private val gitHubUrlPattern = s"""http(s)?://github.com/[$identifier]+/[$identifier]+(/)?""".r.pattern
+  private val readmeUrl = "https://raw.githubusercontent.com/%s/%s/master/README.md"
```
kashike: Have you tested this with varying capitalisation?
- `readme.md`
- `README.MD`
- `readme.MD`

Additionally, GitHub supports more than just `.md` as a file extension for markdown: https://github.com/github/markup/blob/master/lib/github/markup/markdown.rb#L32

app/models/project/Project.scala
```diff
@@ -446,7 +447,18 @@ case class Project(override val id: Option[Int] = None,
     * @return Project home page
     */
   def homePage: Page = Defined {
-    val page = new Page(this.id.get, Page.HomeName, Page.Template(this.name, Page.HomeMessage), false, -1)
+    var body = Page.HomeMessage
```
kashike, line 468: This should only do it if one is not manually provided, yes? If that is not the behaviour with this PR, it should be.

phase: I'm not quite sure what you're saying. This scrapes the README if a source URL is provided, and if it is from GitHub.

kashike: I'm asking if this will overwrite an existing manually-created README.

phase: When a Project is created, it creates a Home Page that has `Page.HomeMessage` in it. Lines 448 - 451 retrieve the Page from database, passing the `page` variable as a default. This changes the default to the README it scraped from GitHub. It's basically replacing the text from `Page.HomeMessage` with the README. So no, it will not replace the Home Page after it has been edited.

windy1:
Can you please make this optional with a setting please? (and during project creation)

app/util/GitHubUtil.scala
phase: This currently only supports `README.md`. I'll update it with different types soon.

phase:
Is there a use case for it being optional? If you're linking your project to GitHub, I can't think of any reason you wouldn't want your README copied. And even if they don't want it, they can easily delete it and replace it with a different home page if they so desire.

progwml6:
@phase if we want to do this we need both a setting for the project to enable/disable github README syncing, and to consider using commonmark for the markdown rendering which appears to be supported by the current library

parlough:
We should be using CommonMark anyway as I believe Discourse has switched to it in its latest few releases.

progwml6:
if discourse is using it move everything to commonmark based rendering for markdown on ore

simon816:
Ore is already using the [flexmark](https://github.com/vsch/flexmark-java) library which implements CommonMark, so nothing needs to be done there.

> if we want to do this we need both a setting for the project to enable/disable github README syncing

This PR is just scraping the README on project creation, github syncing can come later (IMO)

progwml6:
commonmark is one of the rendering options, and not what we are using iirc

ryantheleach:
Some issues from the old OCD list that I think are relevant to review on this PR:

- [ ] By default syntax highlighting is enabled for code blocks, IMO it should be disabled in favor of using the syntax of Github Flavored Markdown ` ```Java ` for example.
- [ ] Inserting images via HTML is not supported, not sure if this is something that would be supported but I thought I'd mention it. 

If they arn't covered by this PR / CommonMark please recreate them or make an explicit comment declining to implement them, as I don't believe they belonged on a minor issue list.

progwml6:
as far as i'm concerned this PR is for GH readme scraping and any commonmark fixes that we deem necessary while we are at it

progwml6:
@phase this pr has conflicts

progwml6:
@phase this pr has conflicts

phase:
Conflicts have been fixed & it will now get any README using GitHub's API.

Katrix:
See #729

PR #290, opened 2017-10-28
Add PGP Display Button on User Pages

phase:
PGP keys can now be displayed on users that have them. This displays the _entire_ PGP key. If this isn't wanted just let me know.

Closes #262

PR #293, opened 2017-10-31
Show PGP key to everyone, and show raw key at all times

### LINKED ISSUE #262: Add a button to view public key
Maybe add a setting to turn on visibility of your key and have it off by default. But this should definitely be implemented. This allows people to verify signatures of files that weren't distributed via Ore, e.g. privately contracted plugins.

Katrix:
Allows everyone to see PGP keys at all times, including info related to the key such as when it was created and so on. This also allows everyone, including the current user to view the raw key.

Here's a preview from a logged out users perspective.

image

Closes #262

### DIFF
--- a/app/views/users/view.scala.html
+++ b/app/views/users/view.scala.html
@@ -92,15 +92,17 @@
                                     <i class="fa @lockIcon action-lock-account" data-toggle="tooltip"
                                        data-placement="top" title="@messages("user.lockAccount")"></i>
                                 </span>
+                            }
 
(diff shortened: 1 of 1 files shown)


kashike:
#290?

Katrix:
Oh sorry. Didn't see that one. Have had this one lying around for a few weeks now. I guess you could say that they are different in the information they show. This one also formats the raw key as left centered instead of justified.

windy1:
Closing because duplicate.

PR #296, opened 2017-11-24
Redirect to Project Page after Download Acceptance (Closes #220)

### LINKED ISSUE #220: Automatic redirect to the plugin page after download acceptance
I think the modal should be dismissed after the user has confirmed the download of the plugin (as the title says).

phase:
I have no idea why the timeout is needed, but it didn't work without it.

Tested on FireFox & Chrome.

Closes #220

### DIFF
--- a/app/views/projects/versions/unsafeDownload.scala.html
+++ b/app/views/projects/versions/unsafeDownload.scala.html
@@ -4,6 +4,7 @@
 @import models.project.Version
 @import models.project.Project
 @import ore.project.io.DownloadTypes.DownloadType
+@import security.NonceFilter._
 @import views.html.helper.CSRF
 @(project: Project,
   target: Version,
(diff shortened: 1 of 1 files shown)


simon816:
> I have no idea why the timeout is needed, but it didn't work without it.

The form's action sends the browser to a new location (the file to download) so by redirecting immediately the browser will cancel that request in favour of taking you back to the project page.

So to implement this behaviour you have to do some hacky work around. see here: https://stackoverflow.com/questions/822707/php-generate-file-for-download-then-redirect/822732#822732

You could potentially do the file download with AJAX but that's a pain in itself.

phase:
Those sound like hacks that wouldn't be as clean as the hack that is already in place.

felixoi:
Seems too hacky for me. Don't want to implement this like it's done here, maybe using another approach later.

PR #302, opened 2017-12-30
Show who Signed a Plugin in the Versions List (Closes #300)

### LINKED ISSUE #300: Show who signed a plugin.
As a user, I'd like to be able to see who uploaded and signed a plugin release.

Whether that be a user on Ore, an Admin, or a shared key / organization key.

phase:
Since Projects can have multiple authors, I've added an authorId field
to Versions (and updated the DB). When someone publishes a Version,
they are assigned as the author.

A key appears in the Versions list next to Versions that have a proper
`authorId`. If they don't have an `authorId` (like all the ones currently
in the DB), there won't be a key. The key links to author's page.

Closes #300

### DIFF
--- a/app/controllers/project/Versions.scala
+++ b/app/controllers/project/Versions.scala
@@ -198,8 +198,10 @@ class Versions @Inject()(stats: StatTracker,
             try {
               this.factory.processSubsequentPluginUpload(uploadData, user, request.project).fold(
                 err => Redirect(call).withError(err),
-                version =>
+                version => {
+                  version.underlying.authorId = user.id.getOrElse(-1)
                   Redirect(self.showCreatorWithMeta(request.project.ownerName, slug, version.underlying.versionString))
+                }
               )
             } catch {
               case e: InvalidPluginFileException =>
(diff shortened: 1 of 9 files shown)


me4502 (approved):

app/views/projects/versions/list.scala.html
```diff
@@ -91,9 +92,16 @@ <h4 class="panel-title pull-left">@messages("project.versions")</h2>
                                                         <span class="date">@prettifyDate(version.createdAt.get)</span>
                                                         @defining(version.downloadCount) { d: Int =>
                                                             <span class="date">
-                                                                @d Download@if(d != 1) { s } else { &nbsp; }
```
phase, line 94: This is a fix for something dumb I did in #289. Extra whitespace was never needed.

app/controllers/project/Versions.scala
```diff
@@ -198,8 +198,10 @@ class Versions @Inject()(stats: StatTracker,
             try {
               this.factory.processSubsequentPluginUpload(uploadData, user, request.project).fold(
                 err => Redirect(call).withError(err),
-                version =>
+                version => {
+                  version.underlying.authorId = user.id.getOrElse(-1)
```
phase, line 202: This assigns the uploader as the author of the version. This is the core line of the PR. The rest of it is adding `authorId` to the DB.

PR #304, opened 2017-12-30
Fix & Clean up Organization Membership Modifications (Fixes #301)

### LINKED ISSUE #301: Admin role for organisations doesn't persist
Browser & Version: `Unknown`

Operating System: `Unknown`

Error message (if applicable): `None`

Steps to reproduce:

I'm currently `Developer` in the `EngineHub` organisation. Ever since the `Admin` rank was added, both the owner of the organisation and numerous people with global access on Ore have tried setting my rank to Admin. This has not applied each time.

* Try setting me to Admin on the EngineHub organisation
* It won't apply
* I'll still be developer

phase:
There was some whitespace being passed into an equals check, causing
it to fail.

Members that have been invited will now display what role they will be,
and that role is editable.

An NPE with Org PGP Keys was fixed by checking to see if it exists
before using it.

The modification icons were spaced out and the text was extracted into `conf/messages`
It looks a lot better now.

Fixes #301

### DIFF
--- a/app/views/users/memberList.scala.html
+++ b/app/views/users/memberList.scala.html
@@ -60,7 +60,8 @@
                                 aria-label="@messages("general.close")">
                             <span aria-hidden="true">&times;</span>
                         </button>
-                        <h4 class="modal-title" id="label-user-delete">@messages("project.removeMember")</h4>
+                        <h4 class="modal-title" id="label-use
+                        r-delete">@messages("project.removeMember")</h4>
                     </div>
                     <div class="modal-body">@messages("project.removeMember.confirm")</div>
                     <div class="modal-footer">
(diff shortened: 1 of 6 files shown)


phase:
There's one more thing I want to fix.

phase:
When editing a member, the role that is selected was set to the top, `Admin`. It will now be set to whatever the rank of the member is.

Should be good to merge!

me4502 (approved):

app/views/users/memberList.scala.html
```diff
@@ -60,7 +60,8 @@
                                 aria-label="@messages("general.close")">
                             <span aria-hidden="true">&times;</span>
                         </button>
-                        <h4 class="modal-title" id="label-user-delete">@messages("project.removeMember")</h4>
+                        <h4 class="modal-title" id="label-use
+                        r-delete">@messages("project.removeMember")</h4>
```
kashike: make this one line again

kashike (changes requested):

app/views/users/memberList.scala.html
phase: Whoops, don't know how that typo got in there.

phase: Typo fixed.

kashike (approved):

Issue #307, opened 2017-12-30
WIP Tags


vectrixdevelops:
Based on what Owen said in the SoS 15 stream, that the work in progress plugins section will be removed from the Forums and replaced by a tag that can be added to plugins on Ore, a long with a warning at download.

phase:
Currently, there are two "tags" that are displayed with plugins: _Sponge_ and _Forge_. These tags contain the version of the software they they are compatible with. (There are [two plugins](https://ore.spongepowered.org/?page=1&platform=Forge) in the current DB that contain a _Forge_ tag, but for some reason their version number is null.)

What's interesting about this is that these "tags" aren't _tags_, they're [boolean fields in the Project class](https://github.com/SpongePowered/Ore/blob/a457726575afaf42082c133fa2d5ec2c4f5b1c43/app/models/project/Project.scala#L59-L60).
```scala
private var _isSpongePlugin: Boolean = false,
private var _isForgeMod: Boolean = false,
```

What I propose is that we remove these and put in an actual _tags_ system that we can use to label plugins with. (These tags should have consistent artwork with the badges I described in #306.) I think if we're going to implement this, we need to have some use cases. Here are some tags that come to mind (with the data they may store):

* Sponge (Version)
* Forge (Version)
* WIP
* Approved (person who approved)
* Sponsored (Sponsor)? (#120)

Anyone got any other ideas for tags? Does this sound like a good system?

Issue #294, opened 2017-11-03
Filter by Platform Version and Minecraft Version


n00b42:
- Filter by Minecraft Version
- Filter by Sponge API Version

ryantheleach:
To effectively add this I think we need more support for specifying a range of versions that a release will work on.

I'd suggest filtering by Sponge API __only__.

 Minecraft Version, if added, can just be a parent category of Sponge API version if we go down that route.

phase:
This should be included with the Tag system I described in #307.

bloodmc:
Yes this needs to be done as the average user will have no idea what Sponge API version means. It's ok for it to be included somewhere but the main search should be by MC version.

phase:
@ryantheleach said

> It should be Minor / Patch version aware, so if a plugin is uploaded with a 4.1 dependency, that would not show on a 4.0.0 filter.
> 
> However a 4.0.0 plugin would show on a 4.1 filter.

:+1: on that

he also offered this mockup for a _first time user_

![](https://user-images.githubusercontent.com/141554/35490294-aa019f96-04ee-11e8-92f9-eb2f09da2aea.png)

phase:
Minecraft Versions are waiting on https://github.com/SpongePowered/DownloadIndexer/issues/7

phase:
I'm adding SpongeForge, SpongeVanilla, & SpongeCommon filtering to this issue too.
See the discussion in https://github.com/SpongePowered/SpongeForge/issues/2367

gabizou:
Will be able to filter by sponge implementation versions too, if a plugin depends on a specific sponge common (to work on sponge vanilla or sponge forge).

phase:
~~changing the title again just for the fun of it~~

Changing the scope of this issue ~~again~~ to encompass all platform version filters, as I will do them all at the same time. Every platform will have an associated version parser. 

Long version strings will be shortened. Things like `-SNAPSHOT` are long and foreign to users. I think we should replace `8.0-SNAPSHOT` with `8.0*` or something similar (maybe the star can be blue or some different color?) to differentiate it from a normal version. Same goes for the new Sponge implementations' `RC`.

| Platform | Version Pattern| Version Example | Notes|
|------------------|----| --|--|
| Sponge API | `MAJOR.MINOR(.PATCH)?(-SNAPSHOT)?` | `8.0` | Patch won't exist for 8.0 and onward. |
[...]

Issue #213, opened 2017-01-27
Multiple Recommended Versions


Katrix:
It would be neat if there was support for displaying the recommended version for more than one version of SpongeAPI (or minecraft in the case of mods). 

I don't think it's that important for sponge plugins as the goal will be to minimize API breakage in the long run from what I understand. However for Forge mods, which can normally only be used on a few or one specific version of minecraft, it's a big hassle to have to try to find the recommended version for your minecraft version.

The way I see it is that for Sponge, it would group them by the api, while for Forge it would just group them after the minecraft version.

---

Edit by @phase: See the current proposal [here](https://github.com/SpongePowered/Ore/issues/213#issuecomment-422453463).

phase:
I'd like to make recommended versions per-_channel_ instead of per-_verision_.

This will allow us to combine tags for projects to display on the front page. I've got an algorithm in mind to sort them.

mbax:
Make sure to preserve disallowing recommended for a channel marked as non-reviewed.

phase:
I would still like to hear input on this.

@dualspiral brought up the idea of having multiple _jars_ per _version_. I think this will unnecessarily complicate things. Having one jar per version means we can separate versions from each other.

Let's say I have a plugin called `PhaseHome`, and I have Sponge API 7, 8, and BungeeCoord specific variants. Let's also say that the differences between API 7 and 8 are too great to be able to put in one jar, so we need these in separate jars. That's 3 jars per version.

Our current way to display this is as such:

| Name | Channel | Tags |
|-----|-|-|
| 1.0.1-bungee  | Bungee  | BungeeCord
| 1.0.1-sponge8 | Sponge8 | Sponge: 8
| 1.0.1-sponge7 | Sponge7 | Sponge: 7
| 1.0.0-bungee  | Bungee  | BungeeCord
| 1.0.0-sponge8 | Sponge8 | Sponge: 8
| 1.0.0-sponge7 | Sponge7 | Sponge: 7

The channels here can be anything the user wants. We could put both the Sponge versions in one channel, but if I were making this plugin this is how I would do it.

[...]

PR #308, opened 2017-12-31
Tag System

phase:
I'm starting to implement the system I described in #307. 

`isSpongePlugin` and `isForgeMod` have been converted into Tags.
Other Tags will exist later, but for now we only have Sponge and Forge Tags.

Each Project contains a Tags String, which will contain a comma-separated list of ints.
These ints refer to Tags in the `project_tags` table. These tags are project-specific, to make
it more extensible and because projects use different versions of the SpongeAPI/ForgeAPI.
The `project_tags` table consists of a `name`, `data`, and `color`. For API tags, the name
and color tags will be the same, and the data tag will refer to the version of the API.

The toughest part will be migrating the current database to fit this system. That's what `79.sql`
(or `80.sql` once #302 gets merged) will do. Currently, it's a mess of SQL code. IntelliJ doesn't
report any issues with it, but I still need to do some testing.

Reviews & comments are always welcome (someone should probably double check my SQL).
This will conflict with #302 once it is merged.


phase:
Wait, storing a list of integers as a string is dumb. Let me fix that.

phase:
The SQL for converting to the new system works! Now to do the reversion code.

phase:
@Zerthick had the brilliant idea to make tags _per version_ instead of _per project_. Plugin Versions will use different API versions, and displaying what API works with what version will be helpful to users.

Zerthick:
So this may be a given, but having a system to filter projects by the tags associated with their recommended version would also be very helpful, if for example I'm only interested in viewing plugins with a recommended version compatible with the version of sponge I am running. I think in that case #213 will probably need to be revisited as it makes sense that plugins my have a separate recommended build for each version of the API

phase:
Filters will be added once the tags are in place. Filtering out WIP projects and versions you don't want will be useful. (#294)

phase:
Now we're getting somewhere! Tags are _per version_. The SpongeAPI/ForgeAPI tags that are being added need to be reviewed, as the way I'm adding the Tags to the DB is weird. The regression SQL also still needs to be written.

phase:
Next major rewrite I need to do is to allow tags to support _multiple_ versions.

Currently, each tag is linked to one project version. If we have 10 versions for 100 projects that are all using Sponge API 6.0.0, that's 1,000 rows in the DB instead of 1. This can be fixed by changing `version_id int` to `version_ids int []`.

phase:
The last thing that needs to be written is the SQL regression.

phase:
Tags migrate perfectly & the regression script is there if we want to revert it.
Filters still need to be done, but that can be done in another PR.

phase:
This needs some testing. I already had @progwml6 run the big query script to create the tags, but I need to test some stuff with Forge still.

On the frontend, I want to have the tags be displayed within the versions list. Shouldn't be too difficult.

phase:
I've tested everything with a copy of the prod DB and it works! :tada: :tada:

Sponge version tags will be displayed in the Projects version list:

example image

phit:
maybe put the minecraft version in brackets? most random people may not necessarily know what api version targets what minecraft version

phase:
I don't think I want to do MC versions in this PR. That can be put in place later, along with other tags. I just want to get the core Tags system in place.

Filtering is another big thing that needs to get completed, and sorting by Sponge/Forge/MC version will be in that system.

phase:
This has conflicts with #316.

phase:
WIP Tags are in! ~~Everything should be completed for this PR!~~ :tada: :tada: :tada: :tada:

wip

phase:
Todo List:
 * [x] Displaying the RV's tags in the plugin list
 * [ ] Fix filtering

phase:
I'm going to do filtering later. There's a lot that needs to be done and it isn't fit for this PR. For now the platform filter will display every project, which shouldn't be a big deal considering only 2 projects are Forge mods in the prod DB.

So I guess this is good to be merged! :tada: :tada: :tada: :tada:

home page

#307 can be closed when this is merged in.

kashike (approved):

felixoi (approved):
I can confirm that's working on my side.

dualspiral:
Putting my user/UX hat on:

I have a question about the WIP tags - but this is more semantics and naming so this could rumble on later after this is merged if desired. I suspect "WIP" is only used because of our "WIP Plugins" forum category - the use case there is a new plugin that has yet to be fully released as a first version. Would WIP be appropriate for established plugins that aren't a work in progress, but want to provide pre-release versions? Should it be more general, something like "pre-release" or "unstable"? 

From a translation point of view, while it's obvious to many, particularly developers, what WIP means, I don't think it's universal, anyone using a translator would not necessarily get the acronym.

Looking at the function, there is a "WIP" checkbox, with no indication as to what it could do. I'd like to see the checkbox have a line of description with it, at least, that explains what that means. If it's left to say "WIP", it explains nothing to new developers. It's a checkbox with three letters. If it had a line that said "This marks this version as a 'work in progress'/'pre-release' and will warn users that the plugin is potentially unstable on download" or something, that gives understanding to the checkbox.

(Also, again, could be a simple addon later down the road, would it be worth adding an "unreviewed" tag for those plugins that have not been reviewed internally yet?)

ryantheleach:
I think WIP should differ from a nightly or pre-release channel, in that it hides by default on the plugin browse / category pages, and only shows if you are looking for WIP plugins specifically.

That's the point of the WIP forums, to gauge interest for a project that may not even have downloads available yet, or may be abandoned before ever releasing, or just to show off stuff to gain interest and motivation.

At the moment Ore is incapable of creating a project without an initial upload(to my knowledge), factoring a WIP project to properly represent what the forum WIP section is will be somewhat painful.
 

Although, duals comment and mine probably should be on https://github.com/SpongePowered/Ore/issues/307

phase:
@dualspiral
I like `unstable`. The WIP name came from #307 (~~so blame @Owexz~~).

Does `Mark this version as unstable` sound good?

We **need** user hats on. As a developer, I know exactly what _WIP_ means when I see it, but sometimes I'm not thinking in the mind of the average Minecraft player (I try my best, but sometimes it slips).

Other tags will come later. I don't want to add anything else to this PR.

@ryantheleach 
When I get filtering implemented, I'm planning on filtering the front page with WIP projects hidden by default.

If projects are on the WIP forum section _without a buildable version_ then I don't really know what they're doing there. Ore shouldn't be a place for _planning_ plugins, it should be a place for _releasing_ plugins. You can plan plugins on the forums. (And allowing projects to be created without a version would break _everything_ and that's a lot of code I don't want to touch.)


<br>

If anyone has any suggestions for colors let me know.

ryantheleach:
That's fine, I just figured I would highlight a usecase of the forums that Ore can't fulfill easily.

dualspiral:
> Does Mark this version as unstable sound good?

Yes, that's clearer. I still think there needs to be a description somewhere as to what "unstable" means on Ore, such as are they hidden by default, for example (as you suggest) - but I think the function above and beyond a tag is outside the scope of this PR, as correctly noted.

phase:
`WIP` Tags are now `Unstable`.
  Any other suggestions?

phase:
:tada:

PR #310, opened 2018-01-01
Redesign of panels and sponsor blocks 

felixoi:
I want to improve the design of ore a bit. Therefore, I started to improve the design of panels and sponsor blocks.

To improve the design of the sponsor blocks I need to know if there are sponsoring conditions which specify the position of the sponsor blocks (top, bottom..).

Please give me feedback, so that I can make even more improvements!


Dockter:
At the moment the current spongepowered.org page has the correct order based on donation amount.  I had toyed with the idea if Ore had the ability to rotated ads or sponsors to have the higher paying donors rotating in more than others.

felixoi:
Ok, but why is Creeperhost featured currently on Ore?

phase:
This looks clean! This sponsor info is in [`conf/application.conf`](https://github.com/SpongePowered/Ore/blob/723dc4b601d544c5d05cb6ff23ada6ae1d6cba16/conf/application.conf.template#L183-L188). It hasn't been changed in [over a year](https://github.com/SpongePowered/Ore/commit/65e63549933434e82845a1ad2f2bfc8d028d8afe), so it probably needs some updating. A web API on the main site to retrieve this information would be the best solution.

felixoi:
@Dockter Are we somehow able to get all brand images with black font color? It's very hard to get a fitting background color for all sponsors..

Dockter:
Sponsors get a bit itchy when you change their logo's.  "Brand recognition".  That being said, a simple text color change from white to black shouldn't be an issue.  Someone good with photoshop in our community can probably easily do this.

felixoi:
@Dockter Implemented now your suggestion. The sequence given in `application.json` is sorted from high to low donation amounts. The higher a sponsor is listed, the greater the chance that it will be displayed.

felixoi:
Someone has more doubts? I think this are all changes I want to make in this PR.

preview

phase:
`public/images/sponsors/multiplay.png` can be removed.

felixoi:
Should be ready to merge now.

phase:
If you're going to change the border color for the categories, you might want to change the border color for the plugin list as well. The light grey won't match the dark grey.

felixoi:
@phase It's on my todo list for another PR. I don't want to put all changes in one PR, it would be too big.
Since it's looking ok with my new changes, it's ok to leave it at the moment.

app/views/home.scala.html
```diff
@@ -5,25 +5,45 @@
...
+
+    var currentIndex = 1
+    rawLogos.foreach(logo => {
+        val count: Int = rawLogos.size - rawLogos.indexOf(logo)
+
+        for(i <- 1 until count + 1) {
+            logos.+=:(logo)
+        }
+
+        println(logo + " - " + count)
```
phase: Can you remove this debug statement?

app/views/home.scala.html
```diff
@@ -5,25 +5,45 @@
...
-    val urls = config.sponge.getStringList("sponsors.urls").get
+    val rawLogos: List[Configuration] = config.sponge.getConfigList("sponsors").get.toList
+    val logos: mutable.MutableList[Configuration] = mutable.MutableList()
+
+    var currentIndex = 1
+    rawLogos.foreach(logo => {
+        val count: Int = rawLogos.size - rawLogos.indexOf(logo)
+
+        for(i <- 1 until count + 1) {
+            logos.+=:(logo)
```
phase: This can be `logos += logo`

parlough (approved):

phase (changes requested):

app/views/home.scala.html
```diff
@@ -5,25 +5,45 @@
...
 
 @randomSponsor = @{
-    val logos = config.sponge.getStringList("sponsors.logos").get
-    val urls = config.sponge.getStringList("sponsors.urls").get
+    val rawLogos: List[Configuration] = config.sponge.getConfigList("sponsors").get.toList
+    val logos: mutable.MutableList[Configuration] = mutable.MutableList()
+
+    var currentIndex = 1
+    rawLogos.foreach(logo => {
+        val count: Int = rawLogos.size - rawLogos.indexOf(logo)
```
phase: This will make MC Pro Hosting appear 7 times more than YourKit, and I don't think that's a fair distribution. Instead, we should balance these out by donation amount/sponsorship tier.

phase (changes requested):

app/views/home.scala.html
connorhartley: Actually, the sponsorship images are ordered appropriately for how much they've donated, MC Pro Hosting being the most and YourKit being the least I think. @Dockter will need to confirm this?

phase: They are in the correct _order_, but the chance of them appearing needs to be proportional to their contribution amount.

felixoi: Also also thought about this.. With we could introduce 4 tiers. Gold, Silver, Bronze and Normal or something similar.

phase: Maybe for now we should have them all be equal until we have a system in place to determine contribution distribution.

felixoi:
Removed sponsor display rate for this PR. When @Dockter tells me his needs I will re-add this in a new PR. Can be merged now.

phase (approved):

XakepSDK:
Make search bar open by default.

felixoi:
@XakepSDK Please open a new issue and explain there why you would do that change.

XakepSDK:
@felixoi i don't like 'discussions only for discussions'(idk, how to translate this correctly)
This is just my UX. Search bar and a big button 'search'.
You said, that you need feedback? Search bar is in top panel. This is my feedback.

phase:
If you make an issue we can discuss it there. This PR isn't for that.

phase (approved):

PR #315, opened 2018-01-03
Add Script to Fix File Paths

phase:
Before ecd8100420e9095500ff409ac738948b25c3eed6, some versions were placed in incorrect locations. This script fixes that.

@progwml6 wanted this in the main repo.


progwml6:
Note for deployment: when these builds get deployed, this script will need to be ran to fix the files on disc

PR #316, opened 2018-01-03
Remove Role Visibility

### LINKED ISSUE #256: Cannot add myself as a developer to a project my organization owns
image

I am able to add other developers which are in the organization just fine though. This occurred with my project [Latch](https://ore.spongepowered.org/IchorPowered/Latch).

I feel representing all developers who are working on the project as important, as it allows users to know who they should talk to and who to relate to the project. I do not necessarily think all organization members should show up though as not all are necessarily working on that project. 

This is the result without me in the project:

image

phase:
Hiding the role of a user is pretty useless. If anyone can think of a good use case for hiding your role, we can work around it, but I providing users with all the people working on a project is good.

Fixes #256

### DIFF
--- a/app/db/impl/schema.scala
+++ b/app/db/impl/schema.scala
@@ -259,7 +259,7 @@ class OrganizationMembersTable(tag: Tag) extends AssociativeTable(tag, "organiza
 
 }
 
-trait RoleTable[R <: RoleModel] extends ModelTable[R] with VisibilityColumn[R] {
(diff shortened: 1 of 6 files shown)


kashike (approved):

ryantheleach:
Was it the role that was hidden? or the fact the user was on the project?

phase:
The user was hidden from the member list.

PR #319, opened 2018-01-03
Reworked footer

felixoi:
preview

### DIFF
--- a/app/assets/stylesheets/_footer.scss
+++ b/app/assets/stylesheets/_footer.scss
@@ -1,89 +1,90 @@
+@import "sponge_variables";
 @import 'pallette';
 @import 'utils';
 
-/* footer */
-
 .footer {
-  position: absolute;
-  bottom: 0;
-  width: 100%;
-  min-height: 60px;
-  height: 60px;
-  padding: 0 10px 0 10px;
-  a { outline: none; }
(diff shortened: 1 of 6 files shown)


phase:
Can you post a screenshot with it in context with the rest of the site?

I like where this is going, but I think we're missing some key links like the TOS (which is currently being written) and the support forum.

felixoi:
@phase 

preview

And I can add the TOS when it's done. Support Forum is under "Help needed"

phit:
REST is an acronym that should be fully capitalized. So either "Ore REST API" or personally I think just "API" would be enough, since you are on the Ore site anyway and it doesn't really matter to the devs (at least for this link) that would click on this what kind of API it is.

phase:
I personally think it might be a little too big. We don't want it to distract from the plugin list.

felixoi:
@phit should be fixed now. Look in the initial comment.
@phase Changed button label to "Support Forum" to make it more clear.

Moreover, the footer has to be "so big" if you wanna display the information nicely. It will not distract from the plugin list, since on Ore Production the list is much longer and the footer will be pushed down.

felixoi:
Added ToS placeholder. Should be ready to merge.

phase (approved):

Issue #309, opened 2017-12-31
Setup dependencies using something like docker compose for development


progwml6:
This is a low priority issue at the moment. For use during development it would be useful to lower entry barriers if we had spongeauth, discourse, postgres, and any other needed dependencies setup properly and linked together using a tool like docker-compose so that all one needs to start doing ore development is have a properly functioning JVM and if interested an IDE setup with a provided config that points to the docker-compose(or similar) environment.

progwml6:
For this to work several things are needed
* putting SpongeAuth on docker hub

advice:
* base docker-compose files on spongeauth ones

Instructions should be added on what needs to be ran to generate any necessary credentials, database setup, etc. for using this system.

it should have instructions both running via IDE and using prebuilt jars, etc. for use in a potential staging server

Scripts should be available to seed the database with various types of accounts, users, fake plugins, etc.

progwml6:
As all showstopper bugs have either been closed or are being fixed, i'm moving this to high priority to enable easier testing, and a simpler setup for the staging server

phase:
I found a [Dockerfile](https://github.com/SpongePowered/Ore/blob/c0103a713a1343d09c3bd6c137711bc5fe80bc1f/.deploy/docker/Dockerfile) from a long time ago. It'll probably be useful.

PR #330, opened 2018-01-05
WIP: Add docker-compose setup #309

progwml6:
This is a WIP local setup for use in easing the complexity of setting up ore.
It currently only has a functioning spongeauth setup. I will be working on adding another db to the postgres instance to handle ore, adding instructions for hooking ore into this, and then getting a local discourse linked as well.

Scripts for generating fake accounts, api keys, etc. will be worked on as well


phase:
See #347

PR #347, opened 2018-01-07
WIP: Start on Docker Integration

phase:
SpongeAuth fails to connect to the DB currently but we're getting
there.


Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ubuntu:latest
+
```
progwml6, line 3: use the alpine based openjdk images please we don't need this large of an image

conf/application.conf.template
```diff
@@ -155,9 +155,9 @@ slick.dbs.default {
         driver = "org.postgresql.Driver"
         url = "jdbc:postgresql://localhost/ore"
         url = ${?JDBC_DATABASE_URL}
-        user = "root"
+        user = "spongeauth"
```
progwml6: leave the defaults, i don't know what this will break. the docker container & local users should set this however they want.

progwml6:
a build script should be writtten for this that adds the needed variables for statusz like so, https://github.com/SpongePowered/Ore/blob/1.0.0/.deploy/build.sh#L40

https://github.com/SpongePowered/Ore/blob/1.0.0/.deploy/docker/Dockerfile#L9-L18 need to be added as well for that page to work.

progwml6:
Additionally, there are a number of readme additions and other needed things in the docker-compose.yaml #330 that need to be in here

conf/application.conf.template
simon816: This is Ore's connection to the database, it shouldn't use the same credentials as the spongeauth connection

Dockerfile
vectrixdevelops: `openjdk:8-jdk-alpine`

phase:
Important commands for working with Docker & Compose:

```
docker build . -t ore_app
docker-compose up
```

Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM openjdk:8-jdk-alpine
+
+MAINTAINER Jadon Fowler <jadonflower@gmail.com>
+
+# Install Activator
+RUN apk add curl
+RUN apk add unzip
+RUN curl -O http://downloads.typesafe.com/typesafe-activator/1.3.6/typesafe-activator-1.3.6.zip
```
felixoi: Use `ADD http://downloads.typesafe.com/typesafe-activator/1.3.6/typesafe-activator-1.3.6.zip typesafe-activator-1.3.6.zip` instead.

phase:
https://github.com/phase/Ore/pull/1

PR #359, opened 2018-01-14
Moved project search from header to home view

### LINKED ISSUE #332: Better header
Atm search bar is closed by default. It can be hard to find searchbar for newbie.
I made this issue: https://github.com/SpongePowered/Ore/issues/253
So, it will be better to make searchbar open by default.

felixoi:
Closes #332.

preview

### DIFF
--- a/app/assets/stylesheets/_nav.scss
+++ b/app/assets/stylesheets/_nav.scss
@@ -60,12 +60,6 @@
   }
 }
 
-.project-search {
-  @include size(0, 40px);
-  padding: 3px;
-  overflow: hidden;
-}
-
 .nav-icon {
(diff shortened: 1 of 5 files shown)


Ewoutvans (approved):

parlough:
I did not see this PR originally, sorry for the late response. I am fine with the search bar being there on the main project page, but with this change that's the only place you can use the search. I often use the search looking for projects and similar from other pages, but now I cannot and I have to re-navigate to the home page. :(

phase:
Well that was an oversight... Do we want to have both the bar and the button? Or maybe hide it if the bar is present?

PR #373, opened 2018-01-16
Give Organization Admins their own Trust Level

### LINKED ISSUE #364: Order of members in organization is wrong
Browser & Version: `Chrome Canary - Version 65.0.3321.0 (Official Build) canary (64-bit)`

Operating System: `macOS High Sierra - Version 10.13.3 Beta`

Error message (if applicable): You would expect the owner to be above the admin.

screen shot 2018-01-14 at 4 32 36 pm

Steps to reproduce:

* Go to https://ore.spongepowered.org/IchorPowered
* Look at the organization members
* See how Vectrix as an admin is higher than Meronat as an owner

### LINKED ISSUE #365: Cannot change role of admin even if you are owner of the organization
Browser & Version: `Chrome Canary - Version 65.0.3321.0 (Official Build) canary (64-bit)`

Operating System: `macOS High Sierra - Version 10.13.3 Beta`

Error message (if applicable): I cannot change Vectrix's rank/role or delete him from the organization at all even though I am the owner. At least there is no UI to do so, unlike there is for nighteyes who is only a developer.

screen shot 2018-01-14 at 4 32 36 pm

Steps to reproduce:

* Add an admin to your organization
* See you cannot change their role or delete them

phase:
Org Owners & Admins had the same Trust Level. Owners couldn't modify
Admin positions because they had Absolute Trust. They are now ordered
correctly, and Admins can modify the positions of everyone except the
Owner.

Closes #364
Closes #365

### DIFF
--- a/app/ore/permission/Permission.scala
+++ b/app/ore/permission/Permission.scala
@@ -1,6 +1,6 @@
 package ore.permission
 
-import ore.permission.role.{Absolute, Limited, Standard, Trust}
+import ore.permission.role._
 
 /**
   * Represents a permission for a user to do something in the application.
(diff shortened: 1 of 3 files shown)


simon816:
Wouldn't changing the trust level of `Absolute` from 3 to 4 have some repercussions? There needs to be a way to migrate all users who are level 3 up to level 4.
I tend to define trust values in multiples of 10 (i.e 10, 20, 30) so any further precision can be inserted at a later point.

phase:
Users have their _roles_ stored in the DB by id, but the trust levels of those roles only affect comparisons. Changing the number doesn't warrant the need for any corrections, since every role keeps their same trust level (except `OrganizationAdmin`, which merely got a new one).

me4502 (approved):

felixoi (approved):

PR #374, opened 2018-01-16
Fixed minor style issues

### LINKED ISSUE #371: New OCD Issue: Misc Style Fixes
- [x] Hovering over the buttons in the upper right puts a white box around them except the new project button. It should be given a white background and the plus should change white. _(background on hover was removed)_
- [x] The white box around the search and authors button when you hover is square but when you hover the user icon it is rounded. _(background on hover was removed)_
- [x] I don't really think the plus and avatar buttons need an arrow next to them(look at forums for example). They are a little bit buggy and add clutter.
- [x] While editing members on an organization or project, the name, edit, and trash buttons are very close. Maybe the trash button should come after the edit button as well.
- [ ] Where it says "This project is under review...", I think it would be beneficial to link to a page which includes information about the review process, how it works, who curates it, et cetera. - @Meronat

felixoi:
Should fix #371.

Preview of navigation fixes:

Fixes also:

### DIFF
--- a/app/assets/stylesheets/_nav.scss
+++ b/app/assets/stylesheets/_nav.scss
@@ -46,6 +46,7 @@
 
 .user-dropdown > li {
   position: relative;
+
   .unread {
     bottom: 6px;
     margin-left: 5px;
@@ -54,18 +55,16 @@
 
 .navbar-main {
   .navbar-right > li > a { padding: 0; }
-  .navbar-right {
-    @include padding-vert(9px, 9px);
-    padding-right: 20px;
-  }
 }
 
 .nav-icon {
   @include transition2(background-color, 0.5s);
   cursor: pointer;
-  padding: 7px;
-  margin-right: 5px;
   text-align: center;
+
+  &:not(:last-child) {
+    margin-right: 20px;
+  }
 
   .icon {
     @include transition2(color, 0.5s);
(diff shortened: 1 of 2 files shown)


phase:
Why were those buttons switched? The members list isn't nearly as important as the profile and new buttons. I think it makes sense to keep them together.

parlough:
The size and shape of the members list is larger and allows for a sort of nice visual progression. Having the thin + in the middle looks strange.

parlough:
Maybe move it all off to the right slightly so Spongie aligns closer to the border of the categories window? Maybe not perfectly aligned but closer to it at least.

felixoi:
@Meronat Yes, I'm already working on it.

SimonFlash:
In addition to the visual progression as @Meronat said, I also consider the functionality of the button to become more (for lack of a better term) 'meta' moving towards the right. Both the organization and sponge profile buttons are links to user information (whether about the user themselves or their organizations). The new button seems to be more of a convenience to me than a necessary part of the UI, and thus should be further away. I would likely be checking organization info more than I would be creating one, so it makes sense in that sense too.

It's one of the things that you have a sense for when you see it, but can't easily explain why.

ryantheleach:
It's not a logical grouping but a styling one. The shape and density of the yellow fades as it goes to the right.

felixoi:
I tried to move it more to the right, but I'm not sure if this is what we want. It currently has the same padding to the left as the sponge icon has to the right (on the right navigation side).

parlough:
@felixoi Can we see an image/gif?

felixoi:
I'm not going to move this to the right. It's hard-coded in every navigation on all sponge sites.

Ewoutvans:
Add carot back to the dropdowns

ryantheleach:
>I'm not going to move this to the right. It's hard-coded in every navigation on all sponge sites.

The problem people have is with the right hand margin on the categories.

2 fixes, 
1. make the page smaller by having the categories align instead.
2. make the page larger by having the categories further right, so peoples brain doesn't associate the alignment.

felixoi:
This is not fixable in the current navigation as is uses a custom container width (like on all sites 💩 ).

felixoi:
Should be ready to merge if no one other has further suggestions. I'm not adding back the carots since the forums aren't using them, too.

Ewoutvans (approved):

PR #375, opened 2018-01-17
Add resolved_at to a flag and make comment non-optional

Ewoutvans:
* Enforcing a comment so we don't crash a plugin page when its NULL.
* Add resolved_at for future statistics page
* Add resolved_by


Ewoutvans:
This PR fixes problems that show NULL errors to admins/moderators on the Ore project by setting all NULL comment in the database to a empty string and preventing the creating of new NULL comments

phase (approved):

Issue #329, opened 2018-01-05
Staff-side Improvements


mbax:
* [x] 1.  Approval Queue self-assignment locking
Goal is to avoid two staff reviewing the same thing. Prior to allowing a user to 'approve' a file, require them to click a button to mark the file as being 'under review.' When clicked, checks against latest before updating for the user. (#351)
    * "Are you sure?" prompt to allow override to someone else.
    * Automatically (process on next queue load?) expire lock-out if locked for longer than some configured period of time (30 minutes? hour? Doesn't have to be too short)
    * Cannot mark more than a configurable quantity (5? 10?) of files at once. 

* [x] 2. Display (to staff) who approved a file and when it was approved on the file's page. (#351)

* [x] 3. Display, if present, the approval history of users on a page accessible from their profile. 

* [x] 4. Staff page. List who is staff specifically on Ore, for convenience (easy jump to request number three above). (#529)

* [ ] 5. Approval stats on a page for staff viewing.
    * [x] Approvals per day. (#381)
    * [x] Flags handled per day (#381)
    * [x] Submissions per day. (#381)
    * [x] Flags created per day (#381)
    * [ ] Recent staff activity (selectable time frames like "last week" "last month" or just a "from: to:")

* [x] 6. Ability to make an entire project page unviewable to non-staff (still viewable by members of the project so they can fix problems and flag it for approval). For situations of pages not compliant with the guidelines to the point they need to be taken down

* [x] 7. Ability for staff to view a list (sortable and searchable) of projects in the above unviewable state. 
  
* [x] 8. Staff notes on projects, accessible from the project page. Tracked by user writing them. This way I could say "PM'd 9 Jan 2018 about guidelines violation" or something. (#391)

Ewoutvans:
> 3. Display, if present, the approval history of users on a page accessible from their profile.

> 4. Staff page. List who is staff specifically on Ore, for convenience (easy jump to request number three above).

Public or private (staff only) pages?

mbax:
I'd say both staff only. The staff listing could be considered for being public later, but I figure the listing on the organization is enough for that.

Ewoutvans:
@mbax is number 8 only visible by staff or can the project owner also see them?

kashike:
I'd say staff only.

mbax:
All of these are staff only.

JBYoshi:
I think #429 fixes point 6, although I'm not 100% sure.

mbax:
[...]

PR #381, opened 2018-01-20
Add stats page

Ewoutvans:
Feedback welcome!

Here is a screenshot:
screenshot-2018-1-20 stats

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -23,6 +23,8 @@ import security.spauth.SingleSignOnConsumer
 import util.DataHelper
 import views.{html => views}
 
+import scala.concurrent.Future
+
 /**
   * Main entry point for application.
   */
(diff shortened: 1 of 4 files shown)


ryantheleach:
@Ewoutvans Looking good, is there an associated issue?

progwml6:
this is part of #329

app/views/users/admin/stats.scala.html
```diff
@@ -0,0 +1,128 @@
...
+                            label: "Unsafe Downloads",
+                            backgroundColor: "lightseagreen",
+                            borderColor: "darkseagreen",
+                            fill: false,
+                            data: @Html(Json.stringify(Json.toJson(unsafeDownloads.map(_._1).toList)))
+                        }]
+                    },
+                    options: {
+                        title: {
+                            text: "Reviews"
```
phase: This should be `Downloads`

app/controllers/Application.scala
```diff
@@ -186,4 +188,29 @@ final class Application @Inject()(data: DataHelper,
     Redirect(ShowHome)
   }
 
+  /**
+    * Show stats
+    * @return
+    */
+  def showStats() = (Authenticated andThen PermissionAction[AuthRequest](ViewStats)) { implicit request =>
+    val queryReviews: Future[Seq[(String,String)]] = this.service.DB.db.run(sql"SELECT foo.data, CAST(foo.when AS DATE) FROM (WITH backup AS (SELECT 0,current_date - (INTERVAL '1 day' * generate_series(0, 9)) days) SELECT COALESCE(data.count, 0) AS data, COALESCE(data.when, backup.days) AS when FROM (SELECT COUNT(*), CAST(approved_at AS DATE) AS when FROM project_versions WHERE approved_at IS NOT NULL GROUP BY CAST(approved_at AS DATE)) data FULL OUTER JOIN backup ON data.when = backup.days) AS foo".as[(String, String)])
```
phase: This looks ugly and unmaintainable. You could make one template SQL file and modify it for the different queries.

phase (changes requested):
This looks good except for the SQL queries.

app/controllers/Application.scala
Ewoutvans: maybe we should make it a sql view that would be more suitable in this situation
what do you think is the best way?

simon816: at the very least, break up the query into multiple lines

app/views/users/admin/stats.scala.html
Ewoutvans: This is fixed now but you placed your comment on the wrong line :)

app/controllers/Application.scala
phase: A twirl template might be overcomplicating things. A file with some placeholders would do fine.

kashike (approved):

phase (approved):

progwml6:
This PR has conflicts that need to be resolved

PR #383, opened 2018-01-21
Update slick dependencies

Ewoutvans:
Update to slick
nothing major changed just some fixes and now requires java8


stephan-gh:
It might be easier to update everything in one run, i.e. also update the Play Framework. The latest version has support for Scala 2.12 which is also optimized for Java 8.

Ewoutvans:
We tried it.. its too big of an undertaking to fix
We will probably update the framework when we split the frontend from the backend 
this is just a minor bump of version thats compatible with the current version of the Play Framework

stephan-gh:
@Ewoutvans Hm, were there so many breaking changes? :o

felixoi:
@Minecrell I already tried to update it but there are some breaking changes and it needs a lot of migration.

PR #390, opened 2018-01-29
Fix notification screen and add notification when plugin is approved

Ewoutvans:
- Fix mobile version of notification screen
- Send notification if project has been reviewed 
- Do not send notification if you upload resource it and are watching the resource

### DIFF
--- a/app/controllers/Reviews.scala
+++ b/app/controllers/Reviews.scala
@@ -11,7 +11,10 @@ import db.ModelService
 import db.impl.{ReviewTable, VersionTable}
 import form.OreForms
 import models.admin.{Message, Review}
+import models.user.{Notification, User}
 import ore.permission.ReviewProjects
+import ore.permission.role.Lifted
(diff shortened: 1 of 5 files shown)


felixoi (approved):

app/controllers/Reviews.scala
```diff
@@ -73,6 +76,29 @@ final class Reviews @Inject()(data: DataHelper,
...
+
+            users.foreach(user => user.sendNotification(Notification(
+              originId = request.user.id.get,
+              notificationType = NotificationTypes.ProjectInvite,
+              message = messagesApi("notification.project.reviewed", slug, versionString)
+            )))
+          } else {
+            version.author.get.sendNotification(Notification(
+              originId = request.user.id.get,
+              notificationType = NotificationTypes.ProjectInvite,
```
lol768, line 97: Why not `VersionReviewed`?

PR #400, opened 2018-01-31
Update upload process to show more info

Ewoutvans:
Show dependency:

schermafbeelding 2018-01-31 om 20 07 51
schermafbeelding 2018-01-31 om 20 08 13
schermafbeelding 2018-01-31 om 20 08 53

Fix bug where last screen doesn't show updated description.

### DIFF
--- a/app/ore/project/factory/PendingVersion.scala
+++ b/app/ore/project/factory/PendingVersion.scala
@@ -1,9 +1,10 @@
 package ore.project.factory
 
 import db.impl.access.ProjectBase
-import models.project.{Project, Version}
+import models.project._
 import ore.Cacheable
 import ore.Colors.Color
+import ore.project.Dependency
 import ore.project.io.PluginFile
 import play.api.cache.CacheApi
 import util.PendingAction
@@ -44,4 +45,18 @@ case class PendingVersion(projects: ProjectBase,
 
   override def key: String = this.project.url + '/' + this.underlying.versionString
 
+  def dependenciesAsGhostTags: Seq[Tag] = {
+    var ghostFlags: Seq[Tag] = Seq()
+    for (dependency <-  this.underlying.dependencies) {
+      if (factory.dependencyVersionRegex.pattern.matcher(dependency.version).matches()) {
+        if (dependency.pluginId.equalsIgnoreCase(Dependency.SpongeApiId)) {
+          ghostFlags = ghostFlags :+ Tag(None, List(), "Sponge", dependency.version, TagColors.Sponge)
+        }
+        if (dependency.pluginId.equalsIgnoreCase(Dependency.ForgeId)) {
+          ghostFlags = ghostFlags :+ Tag(None, List(), "Forge", dependency.version, TagColors.Forge)
+        }
+      }
+    }
+    ghostFlags
+  }
 }
--- a/app/ore/project/factory/ProjectFactory.scala
+++ b/app/ore/project/factory/ProjectFactory.scala
@@ -50,6 +50,7 @@ trait ProjectFactory {
   val cacheApi: CacheApi
   val actorSystem: ActorSystem
   val pgp: PGPVerifier = new PGPVerifier
+  val dependencyVersionRegex = "^[0-9a-zA-Z\\.\\,\\[\\]\\(\\)-]+$".r
 
   implicit val messages: MessagesApi
   implicit val config: OreConfig
@@ -331,9 +332,7 @@ trait ProjectFactory {
       if (dependenciesMatchingName.nonEmpty) {
         val dependency = dependenciesMatchingName.head
 
-        val versionRegex = "^[0-9a-zA-Z\\.\\,\\[\\]\\(\\)-]+$".r
-
-        if (!versionRegex.pattern.matcher(dependency.version).matches()) {
+        if (!dependencyVersionRegex.pattern.matcher(dependency.version).matches()) {
           return
         }
 
--- a/app/views/projects/create.scala.html
+++ b/app/views/projects/create.scala.html
@@ -10,6 +10,8 @@ Page used for uploading and creating new projects.
 @import security.NonceFilter._
 @import views.html.helper.{CSRF, form}
 @import ore.permission.CreateProject
+@import models.project.Tag
+@import models.project.TagColors
 @(pending: Option[PendingProject])(implicit messages: Messages, flash: Flash, request: Request[_],
         service: ModelService, config: OreConfig, users: UserBase)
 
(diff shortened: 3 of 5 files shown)


app/models/project/Tag.scala
```diff
@@ -36,6 +36,7 @@ case class Tag(override val id: Option[Int] = None,
 object TagColors extends Enumeration {
 
   // Tag colors
+  val NoDependency = TagColor(-1, "DARKRED", "WHITE")
```
phase: Add a comment here that says this color shouldn't be used in the DB.

app/views/projects/create.scala.html
```diff
@@ -64,6 +66,20 @@
...
                                 </td>
                             </tr>
+                            <tr>
+                                <td colspan="2"><strong>Dependency</strong></td>
+                                <td>
+                                    <div class="pull-right">
+                                    @if(pending.get.pendingVersion.dependenciesAsGhostTags.isEmpty) {
+                                        @projects.tag(Tag(None, List(), "! no dependency found !", "", TagColors.NoDependency))
+                                    } else {
+                                        @for(t <- pending.get.pendingVersion.dependenciesAsGhostTags) {
```
phase: Cache the list of ghost tags

app/views/projects/versions/create.scala.html
```diff
@@ -54,7 +56,11 @@ <h3 class="panel-title">
                                                 @version.description.map { description =>
                                                   @description
                                                 }.getOrElse {
-                                                  <i class="minor">@messages("version.create.noDescription")</i>
+                                                    @project.description.map { description =>
+                                                        @description
```
phase: Does this map do anything?

app/ore/project/factory/PendingVersion.scala
```diff
@@ -44,4 +44,20 @@ case class PendingVersion(projects: ProjectBase,
 
   override def key: String = this.project.url + '/' + this.underlying.versionString
 
+  def dependenciesAsGhostTags: Seq[Tag] = {
+    val versionRegex = "^[0-9a-zA-Z\\.\\,\\[\\]\\(\\)-]+$".r
+
+    var ghostFlags: Seq[Tag] = Seq()
+    for (dependency <-  this.underlying.dependencies) {
+      if (versionRegex.pattern.matcher(dependency.version).matches()) {
+        if (dependency.pluginId.equalsIgnoreCase("spongeapi")) {
```
phase: This constant is in some class already, along with `forge`.

app/views/projects/versions/create.scala.html
```diff
@@ -106,6 +112,20 @@ <h3 class="panel-title">
...
                                             </td>
                                         </tr>
+                                        <tr>
+                                            <td><strong>Dependency</strong></td>
+                                            <td>
+                                                <div class="pull-right">
+                                                @if(pending.get.dependenciesAsGhostTags.isEmpty) {
+                                                    @projects.tag(Tag(None, List(), "! no dependency found !", "", TagColors.NoDependency))
+                                                } else {
+                                                    @for(t <- pending.get.dependenciesAsGhostTags) {
```
phase: Cache the list again here

app/views/projects/versions/create.scala.html
```diff
@@ -106,6 +112,20 @@ <h3 class="panel-title">
                                                 }
                                             </td>
                                         </tr>
+                                        <tr>
+                                            <td><strong>Dependency</strong></td>
+                                            <td>
+                                                <div class="pull-right">
+                                                @if(pending.get.dependenciesAsGhostTags.isEmpty) {
+                                                    @projects.tag(Tag(None, List(), "! no dependency found !", "", TagColors.NoDependency))
```
phase: This message is pretty bad.

phase (changes requested):
I don't like implementing this as a tag in the backend. This could easily be a frontend check. If there's a tag with an id of `-1`, you're using the system improperly. A frontend warning would be fine.

If we allow different kinds of mods in the future, they may not be using any dependencies (such as jar mods).

app/views/projects/versions/create.scala.html
Ewoutvans: Yeah this map is needed because its an `Option[String]` 
So first it will check if you have a version description if not it will fallback to project description or else it will be no description

The reason I changed it like this is because you can edit the message during the upload. But changing it will add it to the project and not to the version, so if you originally upload a plugin with no description the `version.description` is empty and if you change it it will get populated in `project.description`

public/javascripts/pluginUpload.js
```diff
@@ -100,6 +100,11 @@ function addSig(e) {
  */
 
 $(function() {
+    var platformTags = $('#upload-platform-tags');
+    if (platformTags.size() >= 1 && platformTags.children().length == 0) {
+        platformTags.html('<div class="tags"><span style="color: #FFF; background-color: #8b0000; border-color: #8b0000" class="tag">no platform detected</span></div>');
```
phase: `No Platform Detected`

app/views/projects/versions/create.scala.html
phase: But the map itself isn't doing anything to the variable. `f(x) = x` doesn't do anything.

public/javascripts/pluginUpload.js
phase: And can these colors be extracted to somewhere?

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -50,6 +50,7 @@ trait ProjectFactory {
   val cacheApi: CacheApi
   val actorSystem: ActorSystem
   val pgp: PGPVerifier = new PGPVerifier
+  val versionRegex = "^[0-9a-zA-Z\\.\\,\\[\\]\\(\\)-]+$".r
```
phase: Make this `dependencyVersionRegex`, as it could be talking about project versions.

app/ore/project/factory/PendingVersion.scala
```diff
@@ -44,4 +44,18 @@ case class PendingVersion(projects: ProjectBase,
 
   override def key: String = this.project.url + '/' + this.underlying.versionString
 
+  def dependenciesAsGhostTags: Seq[Tag] = {
+    var ghostFlags: Seq[Tag] = Seq()
+    for (dependency <-  this.underlying.dependencies) {
+      if (factory.versionRegex.pattern.matcher(dependency.version).matches()) {
+        if (dependency.pluginId.equalsIgnoreCase("spongeapi")) {
```
phase: The `Dependency` object has these as constants.

phase (changes requested):
This is a lot better than having the tag in the backend.

felixoi (approved):

PR #403, opened 2018-01-31
Fixed edge case in version list pagination

felixoi:


Ewoutvans (approved):

phase (approved):
Seems hacky but okay.

PR #405, opened 2018-02-03
Add a mime-keyed CSP filter

pschichtel:
@Faithcaio asked me to PR this

I backported this from a Play 2.6 project.
The filter has 2 advantages over Play's standard filter:
1. Different Content-Types can receive different a CSP header which is useful to support SVGs with inline style.
2. The configuration of the header is done in a HOCON map instead of a single string

### DIFF
--- a/conf/application.conf.template
+++ b/conf/application.conf.template
@@ -36,6 +36,7 @@ play {
     ws.timeout.connection              =   10000
     filters.headers.contentSecurityPolicy = "default-src 'self'; script-src 'self' '%NONCE-SOURCE%' https://forums.spongepowered.org https://forums-cdn.spongepowered.org https://www.google-analytics.com/analytics.js; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://forums-cdn.spongepowered.org; img-src 'self' https:; font-src 'self' https://fonts.gstatic.com; frame-src https://forums.spongepowered.org; frame-ancestors 'none'; block-all-mixed-content"
     filters.headers.contentSecurityPolicy = ${?CONTENT_SECURITY_POLICY}
+    filters.enabled                   += "filters.MimeKeyedCspFilter"
     filters.csrf.body.bufferSize       =   1000000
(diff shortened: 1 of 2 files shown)


conf/application.conf.template
```diff
@@ -36,6 +36,7 @@ play {
     ws.timeout.connection              =   10000
     filters.headers.contentSecurityPolicy = "default-src 'self'; script-src 'self' '%NONCE-SOURCE%' https://forums.spongepowered.org https://forums-cdn.spongepowered.org https://www.google-analytics.com/analytics.js; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com https://forums-cdn.spongepowered.org; img-src 'self' https:; font-src 'self' https://fonts.gstatic.com; frame-src https://forums.spongepowered.org; frame-ancestors 'none'; block-all-mixed-content"
```
Ewoutvans, line 37: is this not obsolete with this PR?

pschichtel: it is still being applied, but the new filter will overwrite the header. I kept it for reference.

felixoi (approved):

conf/application.conf.template
felixoi: @pschichtel I wanted to modify something in the CSP and I found out that this is not going to be overriden. Maybe because we switched to play 2.6?

pschichtel: The code this is based on is actually from a play 2.6. Has something changed on your configuration? (especially filter settings)

felixoi: It's the config in the master branch. So I don't think so

pschichtel: can you check if it is even getting applied? Maybe it was wrongly migrated to 2.6? The configuration stuff had quite a few changes. Now that you are on 2.6 I could PR the exact version from my project.

felixoi: I'm not sure if it's applied. If you want to go for the PR.

PR #427, opened 2018-02-08
Remove old download entries

Ewoutvans:
Delete download entries older than 30 days


Zidane:
Does this mean that Ore only keeps versions 30 days back? If so, I have to decline this. Space is an easy issue to fix.

progwml6:
no, this has to do with how we store the keys used for unapproved downloads, as wget and curl don't support cookies, we have to store tokens in the database which we are purging after 30 days as to not artificially increase the database size

felixoi (approved):

Issue #413, opened 2018-02-07
Project visibility states


mbax:
For non-standard project states, a notice should be displayed at the top. 

1. Public
    * Visibility: Everyone.
    * Notice: None.
    * Standard state for a project.
2. New
    * Visibility: Everyone.
    * Notice: "This project is new, and the author may not have finished writing the documentation."
      * Extra notice to anyone who can edit project pages: "This notice will expire in 24 hours. Click 'here' to remove it early."
    * Entry point: Create Project.
    * Exit point:
      * Owner clicks button in notice to publish early.
      * Automatically turned over (thanks, cron?) after one day.
3. Needs changes
    * Visibility: Staff and anyone with project page edit access.
    * Notice: "This project requires changes: [changes listed here]"
      * Extra notice to anyone who can edit project pages: "Click here to indicate you have completed these changes"
    * Entry point: Staff clicking "Needs changes" button (from queue or from project page), filling out what needs changing, and submitting.
    * Exit point: Any non-staff who can view can click a button in the notice indicating changes are done.
4. Soft deleted.
    * Visibility: Staff
    * Notice: "Project deleted by [user]: [reason]"
    * Entry point: Owner or staff click delete button, fill in reason for deletion (will show on notice), and submit.
    * Exit point: Staff clicking button to change state, choosing "needs changes" "new" or "public"

"Hide project" presently exists but would go away under this system.

simon816:
How about a "draft" state when you first create a project.
Currently it'll go public with no information, people may want to get the project pages set up before going live.

mbax:
That's a neat idea, @simon816 !

mbax:
Updated main issue with modifications to be more clear, and incorporate draft.

phase:
Should pages in the `draft` state have an Unstable tag?

mbax:
Updated to rename 'draft' to 'new' to avoid confusion. An actual drafting process can be added 1.7.0 or later, and handled via separate ticket.

JBYoshi:
Can this be closed since #429 is merged?

mbax:
@JBYoshi I'm told that's only part one of the code necessary to do this. :)

felixoi:
I think a better name for this is `project status` and not `visibility`.

PR #429, opened 2018-02-11
Add Visibility states

Ewoutvans:
**Updated**
Colors:
- Cyan: New
- Yellow: Needs changes
- Red: Removed
schermafbeelding 2018-02-12 om 11 11 02

schermafbeelding 2018-02-12 om 11 14 59

schermafbeelding 2018-02-11 om 18 21 49

This introduces the states from #413 
Deleting & NeedsChanges will be in a separate PR

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -12,7 +12,7 @@ import db.impl.schema.ProjectSchema
 import db.{ModelFilter, ModelService}
 import form.OreForms
 import models.admin.Review
-import models.project.{Flag, Project, Version}
+import models.project.{Flag, Project, Version, VisibilityTypes}
 import models.user.role._
 import ore.Platforms.Platform
 import ore.permission._
@@ -60,10 +60,15 @@ final class Application @Inject()(data: DataHelper,
       val actions = this.service.getSchema(classOf[ProjectSchema])
 
       val canHideProjects = this.users.current.isDefined && (this.users.current.get can HideProjects in GlobalScope)
-      val visibleFilter: ModelFilter[Project] = if (!canHideProjects)
-        ModelFilter[Project](_.isVisible)
+      var visibleFilter: ModelFilter[Project] = if (!canHideProjects)
+        ModelFilter[Project](_.visibility === VisibilityTypes.Public) +|| ModelFilter[Project](_.visibility === VisibilityTypes.New)
       else
         ModelFilter.Empty
+      if (this.users.current.isDefined) {
+        val currentUser = this.users.current.get
+        visibleFilter = visibleFilter +|| (ModelFilter[Project](_.userId === currentUser.id.get)
+            +&& ModelFilter[Project](_.visibility =!= VisibilityTypes.SoftDelete))
+      }
 
       val pform = platform.flatMap(p => Platforms.values.find(_.name.equalsIgnoreCase(p)).map(_.asInstanceOf[Platform]))
       val platformFilter = pform.map(actions.platformFilter).getOrElse(ModelFilter.Empty)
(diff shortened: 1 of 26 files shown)


conf/evolutions/default/86.sql
```diff
@@ -0,0 +1,9 @@
+# --- !Ups
+ALTER TABLE projects ADD COLUMN visibility INT NOT NUll DEFAULT 1;
+UPDATE projects SET visibility = 5 WHERE is_visible = false;
+ALTER TABLE projects DROP COLUMN is_visible;
+
+# --- !Downs
+ALTER TaBLE projects ADD COLUMN is_visible BOOLEAN default TRUE;
```
phase: typo in `TaBLE`

app/ore/project/ProjectTask.scala
```diff
@@ -0,0 +1,56 @@
...
+    val future = actions.collect(draftFilter.fn, ProjectSortingStrategies.Default, -1, 0)
+    val projects = this.models.await(future).get
+
+    val dayAgo = System.currentTimeMillis() - draftExpire
+
+    projects.foreach(project => {
+      Logger.info(s"Found project: ${project.ownerName}/${project.slug}")
+      val createdAt = project.createdAt.getOrElse(Timestamp.from(Instant.now())).getTime
+      if (createdAt < dayAgo) {
+        Logger.info(s"Project ${project.ownerName}/${project.slug} is still in Draft changing to Public")
```
phase: `Change $owner/$slug visibility from Draft to Public` would be a little clearer.

app/controllers/Application.scala
```diff
@@ -60,10 +60,15 @@ final class Application @Inject()(data: DataHelper,
...
-      val visibleFilter: ModelFilter[Project] = if (!canHideProjects)
-        ModelFilter[Project](_.isVisible)
+      var visibleFilter: ModelFilter[Project] = if (!canHideProjects)
+        ModelFilter[Project](_.visibility === VisibilityTypes.Public)
       else
         ModelFilter.Empty
+      if (this.users.current.isDefined) {
+        val currentUser = this.users.current.get
+        visibleFilter = visibleFilter +|| (ModelFilter[Project](_.userId === currentUser.id.get)
+            +&& ((ModelFilter[Project](_.visibility === VisibilityTypes.NeedsChanges) +|| ModelFilter[Project](_.visibility === VisibilityTypes.Draft))))
```
phase: Doesn't this leave out projects that the user owns that `NeedsApproval`? I think checking for `ModelFilter[Project](_.visibility =!= VisibilityTypes.SoftDelete)` would be cleaner.

phase (changes requested):
This is looking awesome!

dualspiral:
I'd like to suggest a change to the text you have in the red box you have on a draft page. I assume it's user facing - it's written in the third person so looks odd on a message directed at the author. Most users won't see that message because the project is hidden! I'd suggest something like the following might be better (calculating the time remaining before publish):

"This project is hidden for the next xx hours to allow you to add your project's documentation, after which your project will be published automatically. Once you have completed added documentation, you can choose to publish your project early by clicking the "Publish" button."

I'm open to that being rephrased, but that message needs to be author targeted, not third-person targetted.

dualspiral:
...just realised, the draft page isn't going to be hidden, is it? I misread what was in the ore channels and whatnot.

Still, the message should be different for authors and users. If users see that, the first line of what you have is fine. Sorry for any confusion!

Ewoutvans:
no draft is hidden, we changed it after some discussions
but yeah the messages need to change I forgot to do that

mbax:
To avoid confusion, I'd like to see 'draft' renamed to 'new' or 'newproject'. That way, if we decide we want an actual drafting process we can do that later.

felixoi (approved):

PR #434, opened 2018-02-16
Fixed page anchors not pointing to the correct position

### LINKED ISSUE #432: Incorred Header Alignment
When linking to a header, the first few lines are cut off due to the Sponge navigation bar. See the gif below:

Does anyone read these things?

felixoi:
Fixes #432

### DIFF
--- a/public/javascripts/main.js
+++ b/public/javascripts/main.js
@@ -176,6 +176,34 @@ $(function() {
     });
 });
 
+// Fix page anchors which were broken by the fixed top navigation
+
+var scrollToAnchor = function (anchor) {
+    if (anchor) {
+        var target = $("a" + anchor);
+
+        if (target.length) {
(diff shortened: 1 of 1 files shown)


Ewoutvans:
I wouldn't call this a fix, you broke more stuff.
The anchor doesn't get appended to the URL or history so going back is taking you not where you originally where. 
Because you don't append the anchor people will have trouble linking the page to others. 
If you have a URL with anchor it still doesn't fix the issue.

Ewoutvans:
Now its jumping down to go back up to scroll down. Its strange looking

kapture 2018-02-17 at 18 13 07

ryantheleach:
That's strange, wouldn't have expected that considering you are returning false for the link...

Ewoutvans (approved):

Issue #265, opened 2017-05-31
Inconsistent links between Home page and pages/Home


dualspiral:
(Apologies for a badly worded title/report, I'm struggling to think how to summarise it. Hopefully you'll get the gist of what I'm saying!)

When writing the home page for a plugin, you may want to create a link to various sub pages using Markdown - say to link from `<project>/pages/Home` to `<project>/pages/FAQs`, the seasoned markdown writer would write `[FAQs](FAQs)`. However, as the Home page also exists at the project root, when viewed from that page, the link would be broken when viewed there.

To reproduce this:

* Create an Ore project
* Create a new page, call it "FAQs"
* On the home page, enter the markdown `[FAQs](FAQs)` and save it.
* Click on "Home" under pages on the right hand side.
* Click on the FAQs link in the main text body, it correctly links to the FAQs page
* Click on Docs at the top to go to the root project page (without `pages/Home` in the URL) - you should see the home page
* Click on the FAQs link, you'll get a 404

While the workaround is to just put the full URL in, it would be nice if there could be some special handling so that relative links can be resolved one way or another and not break one of the pages.

simon816:
Perhaps it could even support the MediaWiki link format that GitHub Wiki supports (https://help.github.com/articles/adding-links-to-wikis/)
i.e. `[[Link Text|WikiLink]]`

Ewoutvans:
My suggestion is to redirect 
`<project>` to `<project>/pages/Home` 

But I don't know if making URL's to the pages is the best thing to do because the home page gets mirrored to the forum.
Any other idea's on how to fix it?

phase:
Is this still an issue?

Faithcaio:
Looks fixed to me

dualspiral:
It's not fixed. I just tested it on the Nucleus page (which uses absolute links for now) with `[FAQs](FAQs)`, sam behaviour.

PR #443, opened 2018-02-21
Add WikiLinks, Add duplicate pages

### LINKED ISSUE #421: Cannot name two pages with the same heading
When adding a new page as a sub page, it cannot have the same name as another page that is a parent page.

Steps to reproduce:

* Step 1 Add a parent page, say, `Modules`
* Step 2 To that parent page, add `Home`

You'll get the editor for the `Home` page, when you wanted a new page under `Modules/Home`

Ewoutvans:
- Update Flexmark from 0.18.2 --> 0.32.4
- Add WikiLinks `[[Command]]` will guarantee an always working absolute link
- Add `/pages/:parent/:child` use `[[Commands/Admin]]` to get a working link
- Add duplicate Page names (needs some more testing) if not in the same root so 
       you can have 1 `Admin`-page in Commands and 1 `Admin`-page in root but not 2 `Admin`-pages in root

This fixes #421 and #265 

Example:
schermafbeelding 2018-02-21 om 19 12 13


dualspiral:
> Add WikiLinks [[Command]] will guarantee an always working absolute link

Will that work on the home page that gets mirrored to Discourse - that is, will this syntax work on the forum post Ore generates, or would that not get parsed by Discourse?

felixoi (approved):

progwml6:
Before merging, and behavior changes/additions will need to be documented so that users can understand the new behavior, and how it differs from what currently exists, and what backwards compatibility exists now, and what we plan to remove long term.

phase:
Discourse links need to be handled.

felixoi:
Copied from @progwml6 on discord:
> Just leave the links in these as is for now and file an issue to fix them later

PR #448, opened 2018-02-22
Fix #446

### LINKED ISSUE #446: Queue "In Review" bugs
* "Queued by" is presumably missing a username, or is improperly labeled and should just say "Queued"
* Stored time appears off. Might be based on difference between local user clock and Ore clock, as I am UTC-5

image

Ewoutvans:
- Fix the timezone (maybe?)
- Fix the Queued by now showing 'Unknown'

### DIFF
--- a/app/views/users/admin/queue.scala.html
+++ b/app/views/users/admin/queue.scala.html
@@ -10,6 +10,8 @@
 @(underReview: Seq[(Project, Version)], versions: Seq[(Project, Version)])(implicit messages: Messages, request: Request[_], service: ModelService,
         config: OreConfig, users: UserBase)
 
+@import java.sql.Timestamp
+@import java.time.Instant
 @versionRoutes = @{controllers.project.routes.Versions}
 @currentUserId = @{users.current.get.id.get}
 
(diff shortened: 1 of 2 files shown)


app/views/users/admin/queue.scala.html
```diff
@@ -85,11 +89,11 @@ <h3>@messages("queue.review.none")</h3>
                                     @if(currentReview.isDefined) {
                                         @users.get(currentReview.get.userId).get.name
                                         <br>
-                                        <span data-ago="@currentReview.get.createdAt" data-title="started "></span>
+                                        <span data-ago="@currentReview.get.createdAt.getOrElse(Timestamp.from(Instant.now())).getTime" data-title="started "></span>
```
phase, line 92: This prevents a NPE from being thrown, but it doesn't fix the timezone being off.

phase (changes requested):

app/views/users/admin/queue.scala.html
Ewoutvans: I think it does fix it because now its getting the time since Epoch instead of like "2018-02-25 22:31:12.200" but I could be mistaken

phase: It does that only if `createdAt` is null, which should preferably never be null.

felixoi: That's wrong @phase he added .getTime which is called on createdAt If present or otherwise on Instant.now()

phase: Ah, I misread.

phase (approved):

PR #460, opened 2018-02-27
Fix #456

### LINKED ISSUE #456: "Visibility log" display improvements
image

Rather than showing who set the status, it shows who changed away from that status. From this log, it does not show that I was the person to set the project from its initial public state to the 'new' state but rather says that Grinch 'resolved' the state.

Surely it would make more sense to show on the line of the project state who put it there? This would also avoid the issue of an 'unknown' resolver for the current state.

Ewoutvans:


phase (approved):
For future PRs, can you include _what_ was changed in the commit message? A lot of them are `Fix #blah` which aren't very descriptive. This PR is good to go, though. :+1:

ryantheleach:
It's great having the ticket numbers, but maybe just adding the ticket title as well, or a TLDR of the issue

phase:
The boy learned on #463 :clap:

ryantheleach:
Sure, but that doesn't even have the ticket number :S

Having the ticket number in really helps, and GH even links it to appropriate issues, just without that link, it doesn't aid searching in an old git log away from Github.

I was trying to say, both is preferable.

phase:
oh well he's got the ticket number in the commit message :shrimp:

PR #472, opened 2018-03-03
Add post to forums toggle

### LINKED ISSUE #322: Disable Automated Post
Currently, adding a new release generates a post on the plugins Ore thread documenting information about the release.

* Plugins that have multiple builds for a version (such as multiple APIs) create a version for each build (and thus a forum post for each build). This results in multiple posts for each version, most of which contain the same information. [Spam Example](https://forums.spongepowered.org/t/nucleus-the-ultimate-essentials-plugin/14892/168), Deletion Example: Deletion Example
* Limits the customization the developer has over the release. The release messages that I personally post are significantly different than a version description and changelog on GitHub. I would prefer having the ability to draft my own message announcing new versions that is free to use it's own wording as I see fit.
* Results in complex, unwieldy threads. This is especially relevant with multiple API's, but the majority of Ore threads tend to be post after post of "A new version has been uploaded... it is available for download here." This results in the actual discussion about the project being obscured behind the generated release notes (especially for those that include a lot of information in the release).

I would suggest having a checkbox that allows a user to enable or disable the automated sending of these release messages. To take this a step further, it would be very nice to allow the user to edit the release message created such that it is automatically posted when the version is released, but the contents can be edited freely.

Katrix:
Adds a setting to toggle if Ore actions (currently only creating releases) should automatically also create a new forums post.

Also did some housekeeping to ensure that a post is actually created (unless the setting says otherwise) on the forums no matter how the version was created (previously, if a version were created through the API, a post would not be created).

Bit unsure if I covered everything here as this is probably my biggest pull request so far. As I can't test the forums integration, I haven't tested that the toggle work, but I see no reason for it not to.

Closes #322


app/views/projects/settings.scala.html
```diff
@@ -70,7 +70,8 @@ <h4 class="modal-title" style="color:black;">Comment</h4>
                                 source = settings.source,
                                 licenseName = settings.licenseName,
                                 licenseUrl = settings.licenseUrl,
-                                selected = Some(project.category)
+                                selected = Some(project.category),
+                                postReleasesToForum = settings.postReleasesToForum
```
Ewoutvans: this doesn't compile I believe it has to be `settings.postToForums`

conf/evolutions/default/91.sql
```diff
@@ -0,0 +1,5 @@
+# --- !Ups
+ALTER TABLE project_settings ADD COLUMN post_to_forums BOOLEAN default TRUE;
+
+# --- !Downs
+ALTER TABLE project_settings DROP COLUMN post_to_forums;
```
Ewoutvans: Add a newline

app/views/projects/utils/inputSettings.scala.html
```diff
@@ -83,3 +84,18 @@ <h4>@messages("project.settings.license") <i>(@messages("general.optional"))</i>
...
+
+<div class="setting">
+    <div class="setting-description">
+        <h4>@messages("project.settings.postToForum")</h4>
+        <p>@messages("project.settings.postToForum.info")</p>
+    </div>
+    <div class="setting-content">
+        <label>
+            <input @if(postReleasesToForum) { checked } value="true" form="@form" type="checkbox"
+                    class="form-control" id="forums" name="post-to-forums">
```
Ewoutvans: Due to the `form-control` the text is on a new line and it looks strange:
schermafbeelding 2018-03-03 om 18 02 58
I think its better like this
schermafbeelding 2018-03-03 om 18 04 19

Ewoutvans (changes requested):

felixoi:
Does this still adds the first forum post on project creation?

Ewoutvans:
@felixoi as far as I tested yes.

felixoi:
Then I would suggest to rename `postToForums` to `forumSync` or similar.

Katrix:
Thanks for all the suggestions. Pushed some changes with fixes and new and better names.

Ewoutvans:
@Katrix- Looks great! 

May I suggest you could override the 'global setting' when you do an individual upload?
Like for instance I upload for API 5,7,8 and only want 1 message for the latter one on the forums.
Seems nice to have but isn't really specified in the issue

*Some where on this screen*
schermafbeelding 2018-03-03 om 19 53 46

Ewoutvans (approved):

phase (changes requested):
I want the per-release override @Ewoutvans [described](https://github.com/SpongePowered/Ore/pull/472#issuecomment-370171263) before we merge. The rest looks good.

Ewoutvans (approved):

phase (approved):

Issue #392, opened 2018-01-30
Log DB/File Actions


phase:
Admin/mod actions should be logged. A global page for all actions and a per-user page seems like the best idea.

cc @progwml6

progwml6:
Just to clarify: if it requires permissions to accomplish it MUST be logged this includes reviewing, editing, etc.

phase:
Including normal user actions?

mbax:
> Including normal user actions?

* Edit a page
* Upload a file
* Edit any properties of a file
* Edit any properties of a project

Should log at a minimum who did it, when they did it, and with what IP they did it.

PR #490, opened 2018-03-07
Log User Actions (#392)

phase:
Logs have been added for (in order they appeared on GitHub):

* Editing a project page
* Uploading a new project
* Flagging a project
* Uploading a new icon to a project
* Removing an icon from a project
* Removing a user's membership from a project
* Editing project settings
* Renaming a project
* Changing the visibility of a project
* Publishing a project
* Setting a project as `NeedsApproval`
* Deleting a project
  * Soft-deleting a project
* Editing the description of a version
* Setting the Recommended Version of a project to a version
* Approving a version
* Uploading a version
* Publishing a version
* Deleting a version

Actions are stored in a `user_actions_log` table with the schema:

```sql
  CREATE TABLE user_action_log (
    id         SERIAL PRIMARY KEY,
    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
    user_id    INT REFERENCES users,
    address    INET NOT NULL,
    action     TEXT
  );
```

"Logging" any of these actions with print them out to STDOUT with the `info` level, and add them to the table. 

(#392)


app/models/user/UserAction.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import db.impl.UserActionLogTable
+import db.impl.model.OreModel
+import ore.user.UserOwned
+
+case class UserAction(override val id: Option[Int] = None,
+                      override val createdAt: Option[Timestamp] = None,
+                      private val _userId: Int,
+                      private val _action: String
+                     )
```
phase: this also needs to include the IP

phase:
This is good to go. If you have anything else that you want to be logged just ask @mbax

felixoi:
Needs UI.

[https://gyazo.com/de0880224936eedca4432ace0e0a872b](https://gyazo.com/de0880224936eedca4432ace0e0a872b)

felixoi:
There should be an UI for looking into the logs. Admins should see the whole log with the IP address, for Devs/Mods the IP address should be hidden.

mbax:
Viewing the logs could look much like what we have now for activity log, but preferably searchable by user, project, IP address, action type (multiple parameters at once?).

Logs of actions on a project should be accessible from that project.
Logs of actions by a user should be accessible from that user.
Logs of actions on a file should be accessible from that file (and project).

Also, because the table schema described was unclear, action logging should be clear in what happened. Like, not just that a user's membership was removed from a group but what group it was and what membership they previously had in that group prior to removal. Basically, before/after states for things. Except project pages/changelogs which I believe we were hoping to move to a separately tracked system with diffs?

felixoi:
I'm working on this @phase. 

@mbax 
Do you want to log if someone uploads a project which upload never finished?
And I think for example when a moderator changes a projects icon we don't need a before and after or?
This diff for pages and changelogs isn't as easy as you think.

mbax:
> Do you want to log if someone uploads a project which upload never finished?

I think the only useful logs here would be errors (which wouldn't be part of this feature) in the process that prevent the user from completing the process.

> And I think for example when a moderator changes a projects icon we don't need a before and after or?

Isn't that a part of Auth, not Ore? If in Ore, could be useful to log who did it.

> This diff for pages and changelogs isn't as easy as you think.

I think it's a difficult challenge, and doesn't need to be part of the first release. A goal for the future! :D

conf/evolutions/default/92.sql
```diff
@@ -1,11 +1,15 @@
...
-  action     TEXT
-);
+ CREATE TABLE user_action_log (
+   id         SERIAL PRIMARY KEY,
+   created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+   user_id    INT REFERENCES users,
+   address    INET NOT NULL,
+   context    VARCHAR(40) NOT NULL,
+   context_id INT,
+   action     VARCHAR(40) NOT NULL,
```
phase: Why is this only 40 chars?

felixoi: Because it's only a name, it shouldn't be longer than that

phase:
pic of frontend

Backend to frontend communication is done. It doesn't look too pretty. @felixoi I have no idea how to do the pagination so slap that on here.

app/controllers/project/Pages.scala
```diff
@@ -160,15 +160,19 @@ class Pages @Inject()(forms: OreForms,
...
-                  UserActionLogger.log(request.request, UserActionContexts.PROJECT, data.project.id.getOrElse(-1), UserActions.PROJECT_PAGE_EDITED, "", "") //todo data
                   data.project.getOrCreatePage(pageName, parentId, pageData.content)
                 }
               } else {
                 val pageName = pageData.name.getOrElse(parts(0))
                 data.project.getOrCreatePage(pageName, parentId, pageData.content)
               }
               created flatMap { createdPage =>
-                createdPage.setContents(pageData.content.getOrElse(createdPage.contents))
+                if (pageData.content.isDefined) {
```
Katrix, line 166: Any reason for not using fold/getOrElse here?

app/controllers/project/Projects.scala
```diff
@@ -547,7 +548,8 @@ class Projects @Inject()(stats: StatTracker,
         val data = request.data
         val oldName = data.project.name
         this.projects.rename(data.project, newName).map { _ =>
-          UserActionLogger.log(request.request, UserActionContexts.PROJECT, data.project.id.getOrElse(-1), UserActions.PROJECT_RENAME, s"$author/$newName", s"$author/$oldName")
+          service.insert(UserActionLogEntry(None, None, 1, InetString.apply("::1"), GlobalActionContext.Project, 1, UserActions.VersionDeleted, 1, "1", "2"))
```
Katrix: Any reason for apply call here?

app/models/user/UserActions.scala
```diff
@@ -0,0 +1,103 @@
...
+import db.impl.UserActionLogTable
+import db.impl.model.OreModel
+import enumeratum.values.{IntEnum, IntEnumEntry}
+import enumeratum.{EnumEntry, Enum}
+import ore.StatTracker
+import ore.user.UserOwned
+
+case class UserActionLogEntry(override val id: Option[Int] = None,
+                              override val createdAt: Option[Timestamp] = None,
+                              private val _userId: Int,
```
Katrix: Any reason for hiding the values here if they are then exposed again later? The bytecode is AFAIK identical.

felixoi: did not touch this

app/controllers/project/Projects.scala
felixoi: debug statement removed in local changes.

PR #509, opened 2018-03-17
Add Sync_SSO

### LINKED ISSUE #483: Add sync_sso
Add sync_sso.
sync_sso is a url endpoint in Discourse (forum) to update user accounts (username, email...).
We need this feature so if name changes get added in the feature they are automatically handled by Ore.

[More info](https://meta.discourse.org/t/official-single-sign-on-for-discourse-sso/13045#heading--syncronize_records)

Ewoutvans:
Add sync_sso

added 
- `/api/sync_sso`
- added `apikey` to config
- added `secret` to config

Closes #483

### DIFF
--- a/app/controllers/ApiController.scala
+++ b/app/controllers/ApiController.scala
@@ -334,10 +334,7 @@ final class ApiController @Inject()(api: OreRestfulApi,
           val add_groups = query.get("add_groups")
 
           this.users.get(external_id.get.toInt).map { optUser =>
-
-            if (!optUser.isDefined) {
-              BadRequest(Json.obj("errors" -> Seq("User not found")))
-            } else {
+            if (optUser.isDefined) {
               val user = optUser.get
 
               email.foreach(user.setEmail)
@@ -346,12 +343,15 @@ final class ApiController @Inject()(api: OreRestfulApi,
               avatar_url.foreach(user.setAvatarUrl)
               add_groups.foreach(groups =>
                 user.setGlobalRoles(
-                  groups.split(",").flatMap(group => RoleTypes.values.find(_.title == group).map(_.asInstanceOf[RoleType])).toSet[RoleType]
+                    if(groups.trim == "")
+                      Set.empty
+                    else
+                      groups.split(",").map(group => RoleTypes.withInternalName(group)).toSet[RoleType]
                 )
               )
-
-              Ok(Json.obj("status" -> "success"))
             }
+
+            Ok(Json.obj("status" -> "success"))
           }
         }
       }
--- a/app/controllers/ApiController.scala
+++ b/app/controllers/ApiController.scala
@@ -1,7 +1,9 @@
 package controllers
 
-import java.util.UUID
+import java.util.{Base64, UUID}
 
+import akka.http.scaladsl.model.Uri
+import javax.inject.Inject
 import controllers.sugar.Bakery
 import db.ModelService
 import db.impl.OrePostgresDriver.api._
@@ -11,6 +13,8 @@ import javax.inject.Inject
 import models.api.ProjectApiKey
 import models.user.User
 import ore.permission.EditApiKeys
+import ore.permission.role.RoleTypes
+import ore.permission.role.RoleTypes.RoleType
 import ore.project.factory.ProjectFactory
 import ore.project.io.{InvalidPluginFileException, PluginUpload, ProjectFiles}
 import ore.rest.ProjectApiKeyTypes._
@@ -21,6 +25,7 @@ import play.api.i18n.MessagesApi
 import util.StatusZ
 import play.api.libs.json._
 import play.api.mvc._
+import security.CryptoUtils
 import security.spauth.SingleSignOnConsumer
 import slick.lifted.Compiled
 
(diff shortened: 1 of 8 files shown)


app/controllers/ApiController.scala
```diff
@@ -258,4 +261,64 @@ final class ApiController @Inject()(api: OreRestfulApi,
...
+
+        val confApiKey = this.config.security.get[String]("sync_sso.apikey")
+        val confSecret = this.config.security.get[String]("sync_sso.secret")
+
+        if (apiKey != confApiKey) {
+          BadRequest("API Key not valid")
+        } else if (!CryptoUtils.hmac_sha256(confSecret, sso.getBytes("UTF-8")).equals(sig)) {
+          BadRequest("Signature not matched")
+        } else {
+          val decoded = URLDecoder.decode(new String(Base64.getMimeDecoder.decode(sso)), "UTF-8")
```
Katrix: Play (through Akka) has nicer ways of dealing with query strings. Without looking too much into it, something like `Uri.Query(Base64.getMimeDecoder.decode(sso))` (charset is UTF-8 by default) is what you want to go with. Then call toMap on the result of that. From there, no more vars. Also as another tip, Scala provides a `toInt` method on strings to parse them as ints.

conf/application.conf.template
```diff
@@ -66,6 +66,13 @@ security {
         secret = ${?SPONGE_AUTH_SSO_SECRET}
         timeout = 10000ms
     }
+
+    sync_sso {
+        secret = "changeme"
+        secret = ${?SPONGE_AUTH_SSO_SYNC_SECRET}
```
progwml6: we only need one SSO secret, remove this one

conf/application.conf.template
```diff
@@ -66,6 +66,13 @@ security {
...
+        apikey = "changeme"
+        apikey = ${?SPONGE_AUTH_SSO_SYNC_APIKEY}
```
progwml6: call this SPONGE_AUTH_SSO_API_KEY and move it up to the sso object

progwml6 (changes requested):

app/controllers/ApiController.scala
```diff
@@ -258,4 +262,48 @@ final class ApiController @Inject()(api: OreRestfulApi,
...
+
+          val external_id = query.get("external_id")
+          val email = query.get("email")
+          val username = query.get("username")
+          val name = query.get("name")
+          val avatar_url = query.get("avatar_url")
+
+          val optUser = this.users.get(external_id.getOrElse("-1").toInt)
+
+          if (optUser.isEmpty) {
```
Katrix: This statement does nothing in this position. Also not a fan of testing the content in this way. I'd say the best here would be a pattern match on optUser.

app/controllers/ApiController.scala
```diff
@@ -258,4 +262,48 @@ final class ApiController @Inject()(api: OreRestfulApi,
...
+          val external_id = query.get("external_id")
+          val email = query.get("email")
+          val username = query.get("username")
+          val name = query.get("name")
+          val avatar_url = query.get("avatar_url")
+
+          val optUser = this.users.get(external_id.get.toInt)
+
+          if (optUser.isEmpty) {
+            BadRequest(Json.obj("errors" -> Seq("User not found")))
```
Katrix: This statement does not do anything in this position

app/controllers/ApiController.scala
```diff
@@ -258,4 +262,48 @@ final class ApiController @Inject()(api: OreRestfulApi,
...
+          }
+
+          val user = optUser.get
+
+          email.map(user.email_=)
```
Katrix: Prefer foreach for these instead of map

app/controllers/ApiController.scala
```diff
@@ -258,4 +262,48 @@ final class ApiController @Inject()(api: OreRestfulApi,
...
+        val confSecret = this.config.security.get[String]("sso.secret")
+
+        if (apiKey != confApiKey) {
+          BadRequest("API Key not valid")
+        } else if (!CryptoUtils.hmac_sha256(confSecret, sso.getBytes("UTF-8")).equals(sig)) {
+          BadRequest("Signature not matched")
+        } else {
+          val query = Uri.Query(Base64.getMimeDecoder.decode(sso))
+
+          val external_id = query.get("external_id")
```
Katrix: Bit unsure if we instead want to use normal camel case for these values

app/controllers/ApiController.scala
```diff
@@ -258,4 +262,48 @@ final class ApiController @Inject()(api: OreRestfulApi,
...
+        val sig = formData._2
+        val apiKey = formData._3
+
+
+        val confApiKey = this.config.security.get[String]("sso.apikey")
+        val confSecret = this.config.security.get[String]("sso.secret")
+
+        if (apiKey != confApiKey) {
+          BadRequest("API Key not valid")
+        } else if (!CryptoUtils.hmac_sha256(confSecret, sso.getBytes("UTF-8")).equals(sig)) {
```
Katrix: Use `==` instead of equals

app/controllers/ApiController.scala
```diff
@@ -1,7 +1,9 @@
 package controllers
 
-import java.util.UUID
+import java.net.URLDecoder
```
Katrix: Check unused imports

Katrix (changes requested):

PR #511, opened 2018-03-19
Fix Role Clearing

phase:
Role Clearing wasn't taking into account the ID of the org/project that
was being removed.

clearRoles(User) was made an abstract function in MembershipDossier so
that implementation of it could access the specific Table members
needed to check against (Project#membership defines the MembersTable
  to be ProjectMembersTable which contains a projectId to check
  against in its implementation of clearRoles).

Signed-off-by: Jadon Fowler <jadonflower@gmail.com>


felixoi:
Do we have to fix ewout manually now?

Ewoutvans (approved):

PR #516, opened 2018-04-02
Add that non-home pages have a longer max length

### LINKED ISSUE #515: Increase page size limit on non-home Ore pages
In order to make Nucleus compliant with the Ore guidelines, I need to be able to add my configuration files. However, they are larger than the current limit (of what I believe) is 32kB). My configuration files that I would want to add to Ore are my `main.conf`, currently ~53kB and my `commands.conf`, currently ~61kB. (I've also had to contract my commands and permissions pages to make it fit, though in all fairness, those changes aren't negative for me and in the long run would be surpassed by #422.)

Keeping the 32kB limit makes sense for any page that is mirrored to Discourse due to the forum post size limit. However, for other doc pages the limit is possibly arbitary and prevents me from fulfilling the guidelines. Is it possible to raise the limit?

Ewoutvans:
Home pages (and new version descriptions) still have the max length of 32000
Other pages now have a character limit of 75000

Fixes #515

### DIFF
--- a/app/form/OreForms.scala
+++ b/app/form/OreForms.scala
@@ -171,8 +171,16 @@ class OreForms @Inject()(implicit config: OreConfig, factory: ProjectFactory, se
     "parent-id" -> optional(number),
     "name" -> optional(text),
     "content" -> optional(text(
-      maxLength = MaxLength
-    )))(PageSaveForm.apply)(PageSaveForm.unapply))
+      maxLength = MaxLengthPage
+    )))(PageSaveForm.apply)(PageSaveForm.unapply) verifying("error.maxLength", pageSaveForm => {
+      val isHome = pageSaveForm.parentId.isEmpty && pageSaveForm.name.contains(HomeName)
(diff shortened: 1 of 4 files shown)


app/form/OreForms.scala
```diff
@@ -171,8 +171,16 @@ class OreForms @Inject()(implicit config: OreConfig, factory: ProjectFactory, se
     "parent-id" -> optional(number),
     "name" -> optional(text),
     "content" -> optional(text(
-      maxLength = MaxLength
-    )))(PageSaveForm.apply)(PageSaveForm.unapply))
+      maxLength = MaxLengthPage
+    )))(PageSaveForm.apply)(PageSaveForm.unapply) verifying("error.maxLength", pageSaveForm => {
+      val isHome = pageSaveForm.parentId.isEmpty && pageSaveForm.name.getOrElse("") == HomeName
```
Katrix: `pageSaveForm.name.contains(HomeName)`

felixoi (approved):

PR #517, opened 2018-04-06
Improve usage around Future

Katrix:
This PR aims to improve the usage of Future in the current codebase.

Main points include
* Abstract `Future[Option[A]]` and `Future[Either[E, A]]` behind `OptionT[Future, A]` and `EitherT[Future, E, A]`
    * Exists as a combination of their inner and outer type. Most methods found on `Option` and `Either` can also be found on these.
    * Main methods to remember
        * `map`: has type `A => B`
        * `semiFlatMap`: has type `A => Future[B]`
        * `flatMap`: has type `A => OptionT[Future, B]` and `A => EitherT[Future, E, B]`
        * `flatMapF`: has type `A => Future[Option[B]]` and `A => Future[Either[E, B]]`
        * `transform` has type `Option[A] => Option[B]` and `Either[E1, A] => Either[E2, B]`
        * `subFlatMap` has type `A => Option[B]` and `A => Either[E2, B]`
    * Removes many instances of nested map/exists/forall and more
    * Moves error handling out of main method logic
    * In particular, these two patterns have been mostly removed 
```scala
optFoo.flatMap {
  case None => Future.successfull(None)
  case Some(foo) => <map foo>
}
```

```scala
optFoo.flatMap {
  case None => Future.successfull(BadRequest)
  case Some(foo) => <map foo with possibly more BadRequests>
}
```
* Tries to separate error handling from the main logic
* Use more for comprehensions when applicable
* Tries to parallelize stuff, and make parallelization easier
* No more global `ExecutionContext` (the default global one provided by Scala is not the best for a web application)


Katrix:
Alright, I think I'm mostly done with cleaning up the codebase around the use of Future. Now it's on you guys to review the changes.

So, just a few more words on the transformers I've added (`OptionT` and `EitherT`). While I have mostly used them for for `Future` and `Id` (`type Id[A] = A`), there is nothing that prevents their use for other data types as long as the needed implicits are in scope (I've written the needed implicits for the types I think will be the most relevant in the future). In a general sense, they wrap a `F[Option[A]]` and `F[Either[A, B]]` in an `OptionT[F, A]` and `EitherT[F, A, B]`. When `F == Id`, they behave like their normal counterparts. You can also use types like List, Option, Either and more, or nest them. Types like `EitherT[Option, A, B]` and `EitherT[F, E, OptionT[G, A]]` are perfectly valid types.

app/controllers/Users.scala
```diff
@@ -76,17 +77,16 @@ class Users @Inject()(fakeUser: FakeUser,
...
-              this.redirectBack(request.flash.get("url").getOrElse("/"), u)
-            }
-          }
-      }
+      this.sso.authenticate(sso.get, sig.get)(isNonceValid).semiFlatMap { spongeUser =>
+        // Complete authentication
+        for {
+          fromSponge <- User.fromSponge(spongeUser)
+          getOrCreate <- this.users.getOrCreate(fromSponge)
+          pulledForum <- getOrCreate.pullForumData()   //TODO These two pull methods at the moment just return this at the end.
```
felixoi: This doesn't have to be returned. It updates the user data inside the pull function. We can ignore their result IMO.

app/controllers/Organizations.scala
```diff
@@ -90,26 +87,24 @@ class Organizations @Inject()(forms: OreForms,
...
-              role.setAccepted(true)
-              Ok
-            case STATUS_UNACCEPT =>
-              role.setAccepted(false)
-              Ok
-            case _ =>
-              BadRequest
-          }
+    user.organizationRoles.get(id).semiFlatMap { role =>
+      //TODO: Why access the organization when it's only used for one of the statuses?
```
felixoi: That's a bit strange. I think we add the user to the organization before he has accepted this invite. In order to this we have to remove him when he declined and that's why we have to access to the organization here. You are right we only need this when the user declined.

app/models/viewhelper/ProjectData.scala
```diff
@@ -73,44 +76,53 @@ object ProjectData {
       lastVisibilityChange,
       lastVisibilityChangeUser)
 
+    //TODO: Why Future here?
```
felixoi: maybe to have to same usage as `def of[A](project: Project)` but we don't need this here I think

app/ore/rest/OreRestfulApi.scala
```diff
@@ -248,6 +249,7 @@ trait OreRestfulApi {
 
     val limited = filtered.drop(offset.getOrElse(0)).take(lim)
 
+    //TODO: Why is this an Option again?
```
felixoi: Good question 😮 should maybe just return an empty json string or?

phase:
I don't know if this is just me but I'm getting a compilation error.

```
type mismatch;
 found   : scala.concurrent.Future[Option[models.user.User]]
 required: util.functional.OptionT[scala.concurrent.Future,models.user.User]
```

```
In /home/phase/projects/scala/Ore/app/db/impl/access/UserBase.scala:188
184    */
185  def current(implicit session: Request[_], ec: ExecutionContext): OptionT[Future, User] = {
186    OptionT.fromOption[Future](session.cookies.get("_oretoken"))
187      .flatMap(cookie => getSession(cookie.value))
*188     .flatMap(_.user) 
189  }
190
191}
```

IntelliJ isn't showing any errors but Play is.

Katrix:
Did you remember to clean sbt's cache?

phase:
um..yes....

oh hey things magically started compiling!

PR #528, opened 2018-04-15
Default offsets to 0. Resolves #525

### LINKED ISSUE #525: Missing data in API vs website
It seems that the json API is returning outdated data than that listed on the website.

For example:

https://ore.spongepowered.org/Nucleus/Nucleus/versions - there's a version released April 2nd, and April 12th, but...

https://ore.spongepowered.org/api/v1/projects/Nucleus/versions - the latest listed version was released on September 10th.

Is this something I'm misunderstanding with the API perhaps, or is the data not available?

Cherry:
Defaulting to 25 produces unexpected results. See https://github.com/SpongePowered/Ore/issues/525 for more info.

### DIFF
--- a/app/ore/rest/OreRestfulApi.scala
+++ b/app/ore/rest/OreRestfulApi.scala
@@ -71,7 +71,7 @@ trait OreRestfulApi {
 
       filtered sortBy { case (p, _, _) =>
         ordering.fn.apply(p)
-      } drop offset.getOrElse(25) take lim
+      } drop offset.getOrElse(0) take lim
     }
 
     val query = filteredProjects(offset, lim)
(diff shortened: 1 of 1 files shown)


app/ore/rest/OreRestfulApi.scala
```diff
@@ -349,7 +349,7 @@ trait OreRestfulApi {
     */
   def getUserList(limit: Option[Int], offset: Option[Int])(implicit ec: ExecutionContext): Future[JsValue] = {
     for {
-      users <- service.DB.db.run(TableQuery[UserTable].drop(offset.getOrElse(25)).take(limit.getOrElse(25)).result)
+      users <- service.DB.db.run(TableQuery[UserTable].drop(offset.getOrElse(0)).take(limit.getOrElse(0)).result)
```
felixoi: Limit should be 25 by default

Cherry: Whoops, thanks for catching that.

PR #530, opened 2018-04-18
Add version visibilities resolves #479

### LINKED ISSUE #479: Moderator access to deleted versions
- [x] Files deleted should be visible, but clearly marked as deleted, for Ore staff. Should be accessible, downloadable, etc. 
- [ ] Admin should be able to restore deleted files.

Katrix:
Adds version visibilities in the same way we have project visibilities (a lot of the code is shared between them). Deleting versions now soft delete them.

Currently, only the public and soft deleted visibility types are used, although using the others shouldn't be too hard.

Stuff not done
* The admin actions drop-down has a weird placement (help D:)


Katrix:
Ok, so this works now kind of. It's hacky and I'd prefer if we had the 2.0 tokens, but we can change that when we get there.

This is still broken though, and I don't know how to fix it.
image

phase:
Here's a picture of the version list (ripped from Discord):

wow pretty

app/views/projects/versions/view.scala.html
```diff
@@ -115,6 +125,16 @@ <h1 class="pull-left">@v.v.versionString</h1>
                                 </ul>
                             </div>
 
+                            @if(request.hasUser && request.data.globalPerm(ViewLogs)) {
+                                <button class="btn btn-alert dropdown-toggle" type="button" id="admin-version-actions" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
```
Ewoutvans: wrap the `button` and `ul dropdown-menu` around `<div class="dropdown dropdown-menu-right" style="display: inline-block">
` to fix the alignment issue.

Ewoutvans (approved):

PR #531, opened 2018-04-18
Add linter and formatter

Katrix:
Adds two linters (scalac has some stuff built in), and a formatter.

The formatter is opinionated, but I hope that should not be a problem. I'll try to tweak the settings for it to better match the current codebase as currently there are a few things to fix.

Aso for the linters (and scalac options), let's go over the settings one by one and let me explain why they are there.

**Scalac**
* -deprecation Emits deprecation warns and says what is deprecated. We want this.
* -encoding utf-8 All source files should be encoded with UTF-8
* -explaintypes Explains type errors when they occour in more detail. Should hopefully mean fixing compile errors is easier
* -feature Emits feature warnings if a feature flag is not imported
* -unchecked Emits warnings for when the code relies on assumtions
* -Xcheckinit Wraps vals in traits in check methods (checks if they are initialized). Prevents NPE. Techically we shouldn't be using vals in traits at all
* -Xlint Enables all scalac linter options
* -Yno-adapted-args Disables auto tupling of arguments. Makes some code not compile currently.
* -Ywarn-dead-code Finds dead code. Also points out any ??? in the code.
* -Ywarn-infer-any Warns if Scalac infers a type to be any. This should never happen.
* -Ywarn-nullary-override Warns if a method `def foo(): Unit` is overridden as `overridedef foo: Unit`. Hides intent.
* -Ywarn-nullary-unit Warns if a method which returns unit is not defined with `()`. Methods that return unit are only run for their side effects. 
* -Ywarn-unused Emits warnings for unused stuff. (Imports, defs, vals, bound variables in matches), that sort of stuff. 

**Scalafix**
* DisableSyntax
* * keywords = return, null. Both return and null can introduce unexpected code. Null in scala is not recommended, and return can in some cases change the behavior of the program. I'd be willing to bump it down to non local return when a rule like that is added. For more discussion, see the blog post about [why return is harmfull](https://tpolecat.github.io/2014/05/09/return.html) Dotty (scala 3.0) is also considering dropping non local return.
* * DisableSyntax.noTabs Tabs are not used in scala
* * DisableSyntax.noValInAbstract Abstract vals often leads to errors. In most cases, a def is what was meant instead.
* * DisableSyntax.noImplicitConversion Implicit conversions can result in suprising behavior. They are also intended to be limited in Dotty.
* LeakingImplicitClassVal Emits warnings if an implicit AnyVal class is not defined with a private val
* MissingFinal Emits warnings if case classes are not final. Non final case classes unseal a sealed hierarchry
* NoInfer Emits warnings if implicit values don't have explicitly defined types. This is a compile error in Dotty, and s warning in IntelliJ
* ProcedureSyntax Procedure syntax is deprecated, and removed in Dotty.
* RemoveUnusedImports Emits warnings for unused imports

I'd also like to add a CONTRIBUTING.md file which details how to run scalafix, what other rules should be followed, (All public values should have types set is one I'd love to add), as well as a recap of recommended scala practices and style, overlooked Option and collection methods (lot's of code that can be simplified using there), and other misc stuff.

I'd like to say that I look at this mostly as a starting point. I'd for example also like to remove all uses of scala's `Enumratium` (it's a joke. Java's Enum has more features), and either replace it with a raw ADTs, or the library [enumeratum ](https://github.com/lloydmeta/enumeratum) (what Enumratium whishes to be. Also might be included in the upcoming scala playform). I will also not be able to remove all warnings which will be generated from this in this PR alone. I'd instead like to seperate those bigger changes out into seperate PRs.


phase:
I'd like this branch in the repo so we can add to it (for CONTRIBUTING.md, etc).

What about new lines at the end of files? You've currently got two files in this PR that no longer have them.

Katrix:
Not sure if I have write access to the Ore repo. Working on the CONTRIBUTING.md file at the moment. Scalafmt always inserts newlines at the end of files so as long as it's run, that should be fine. It should handle converting line endings to UNIX too (yes, we have windows ones).

Katrix:
BTW, do we have an agreed upon line length for the project? I feel like I've seen everything, so I have no idea.

felixoi:
No we don't have IMO but we should add one maybe.

phase:
I think the standard Java line length is 120(?) which should be fine. IntelliJ complains a _lot_ about "API functions not specifying a return type" which we should fix. The goal should be to open the project and _not have IntelliJ scream at you_.

felixoi:
I hate too Short line length, I would appreciate something higher. Also this intellij warning isn't right at all. You don't have to specify a return type in Scala. You can just disable the inspection for this

Katrix:
True you don't have to, but it's strongly recommend by the Scala style guide. https://docs.scala-lang.org/style/types.html#inference

For line length, I'd say either 120, or 150. Good Scala code should never hit those numbers except in a few cases though.

phase:
I'm going to schedule this for v1.7 as this is pretty low priority.

ryantheleach:
+1 on API **MUST** specify return types. It complains for an extremely good reason. non-api non-public non-interface non-trait methods are fine though.

If you want it more refactor friendly, you can always define type aliases.

phase (commented):
I'm good with the changes in here. The test suite needs to be formatted & linted too.

Things labeled `FUTURE` in the README will be covered in later PRs.

We're going to be turning `scalafix` on when `0.6.0` (or `0.7.0`?) comes out.

We will be waiting for the new warnings introduced to be fixed before merging.

felixoi (approved):

phase (approved):

PR #532, opened 2018-04-18
Redirect /author/slug to /author/slug/pages/Home

### LINKED ISSUE #265: Inconsistent links between Home page and pages/Home

Ewoutvans:
Fixes #265 by redirection /author/slug to /author/slug/pages/Home

### DIFF
--- a/app/controllers/project/Projects.scala
+++ b/app/controllers/project/Projects.scala
@@ -10,7 +10,7 @@ import db.ModelService
 import discourse.OreDiscourseApi
 import form.OreForms
 import javax.inject.Inject
-import models.project.{Note, VisibilityTypes}
+import models.project.{Note, Page, VisibilityTypes}
 import models.user.User
 import ore.permission._
 import ore.permission.scope.GlobalScope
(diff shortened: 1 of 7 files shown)


mbax:
Why can't we just display the content of `/author/slug/pages/Home` at the main `/author/slug` address? This makes for uglier URLs.

simon816:
Agree with @mbax 
IMO it should do something like:
- Prepend `pages/` to all relative links unless it already begins with `pages/`
- Maybe even rewrite links to `pages/Home` back to the project page

phase:
Prepending `pages/` to the relative links won't work on pages that aren't the home page.

I believe this is not the way we want to fix this issue so I will close this PR.

PR #533, opened 2018-04-21
Some Cleanup

felixoi:
- [x] Renamed visibility to project state to fit our current usecase
- [x] Refactored page max length calls
- [ ] Make the messages file great again


app/db/impl/model/common/StateClassifiable.scala
```diff
@@ -0,0 +1,22 @@
...
+
+  override type M <: StateClassifiable { type M = self.M }
+  override type T <: StateColumn[M]
+
+  /**
+    * Returns the current state of the [[Model]].
+    *
+    * @return the current state
+    */
+  def state: ProjectState
```
felixoi, line 20: This is not really a common trait when the type is ProjectState but I don't know what to do here.

Katrix: I make quite a bit of use of this trait in the version visibility (or now state PR)

felixoi: Ok, so I leave it like this for now and you change it then in your PR?

felixoi:
Closing this since it has many merge conflicts and resolving them together with dropping the rename commit ends in hell.

PR #547, opened 2018-05-19
Add delete api keys and invalidate sessions endpoints

Katrix:
Adds two admin endpoints to delete all api keys before a specific data, and invalidate all sessions before a specific date.

To use send a request to the specified endpoint with the session cookie and a body with the key `from` formatted as `YYYY-MM-DD HH:mm:ss`


conf/routes
```diff
@@ -11,6 +11,8 @@ POST    /admin/user/:user/update                                    @controllers
 #GET     /admin/reset                                                @controllers.Application.reset()
 GET     /admin/seed                                                 @controllers.Application.seed(users: Int ?= 200, projects: Int ?= 1, versions: Int ?= 2, channels: Int ?= 2)
 #GET     /admin/migrate                                              @controllers.Application.migrate()
+POST    /admin/deleteApiKeys                                        @controllers.Application.deleteApiKeys()
+POST    /admin/invalidateSesseions                                  @controllers.Application.invalidateSessions()
```
felixoi: Spelling

phase:
I do not like having these as endpoints. We already had a mishap with ‘/admin/seed’, and this just opens a chance for another one. These can easily be made into one time Python scripts like the other migrations.

Katrix:
Already resolved differently.

PR #562, opened 2018-07-09
Replace Option id and timestamp with custom types

Katrix:
Replaces Option ids and timestamps with custom types. This is done as those two values are 99% of the time not optional, and instead something the DB is responsible for filling out. This PR also aims to drop support for data that is in the wrong configuration. If an object id is missing, an exception is generated instead of possibly creating invalid data.

Another addition is the type `ObjectReference`. It's intended to be used for all sorts of object ids. At the moment it's just aliased to Int, but in the future it would be great if it could be made opaque.

For the most part I expect stuff to work as it used to. `Tag` needs to be tested as they have no creation date. `Review` also does some funky stuff with the `createdAt` field so it needs to be tested too. I'll hopefully get around to testing it once I get back to my main computer.

As for the typelevel programming stuff in the schema, I've tried to make it as understandable as possible. Tell me if there is still some stuff that isn't completely clear. Technically Akka ships with a shaded shapeless, but I don't think it's wise to use that as it's probably regarded as internal.

### DIFF
--- a/app/db/dbObjects.scala
+++ b/app/db/dbObjects.scala
@@ -1,0 +1,48 @@
+package db
+
+import java.sql.Timestamp
+
+import db.ObjectId.Uninitialized
+
+sealed trait DbInitialized[A] {
+  def value: A
+  def unsafeToOption: Option[A]
+}
+
+sealed trait ObjectId extends DbInitialized[ObjectReference]
+object ObjectId {
+  case object Uninitialized extends ObjectId {
+    override def value: Nothing = sys.error("Tried to access uninitialized ObjectId")
+    override def unsafeToOption: Option[Nothing] = None
+  }
+
+  private case class RealObjectId(value: ObjectReference) extends ObjectId {
+    override def unsafeToOption: Option[ObjectReference] = Some(value)
+  }
+
+  def apply(id: ObjectReference): ObjectId = RealObjectId(id)
+
+  def unsafeFromOption(option: Option[ObjectReference]): ObjectId = option match {
+    case Some(id) => ObjectId(id)
+    case None => Uninitialized
+  }
+}
+
+sealed trait ObjectTimestamp extends DbInitialized[Timestamp]
+object ObjectTimestamp {
+  case object Uninitialized extends ObjectTimestamp {
+    override def value: Nothing = sys.error("Tried to access uninitialized ObjectTimestamp")
+    override def unsafeToOption: Option[Nothing] = None
+  }
+
+  private case class RealObjectTimestamp(value: Timestamp) extends ObjectTimestamp {
+    override def unsafeToOption: Option[Timestamp] = Some(value)
+  }
+
+  def apply(timestamp: Timestamp): ObjectTimestamp = RealObjectTimestamp(timestamp)
+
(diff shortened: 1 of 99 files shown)


app/db/dbObjects.scala
```diff
@@ -0,0 +1,48 @@
...
+    override def unsafeToOption: Option[Timestamp] = Some(value)
+  }
+
+  def apply(timestamp: Timestamp): ObjectTimestamp = RealObjectTimestamp(timestamp)
+
+  def unsafeFromOption(option: Option[Timestamp]): ObjectTimestamp = option match {
+    case Some(id) => ObjectTimestamp(id)
+    case None => Uninitialized
+  }
+}
```
phase, line 48: No newline

phase (approved):
Bless the lord this is finally happening 🧡

app/ore/project/ProjectTask.scala
```diff
@@ -44,8 +44,8 @@ class ProjectTask @Inject()(models: ModelService, actorSystem: ActorSystem, conf
     val dayAgo = System.currentTimeMillis() - draftExpire
 
     projects.foreach(project => {
-      Logger.debug(s"Found project: ${project.ownerName}/${project.slug}")
-      val createdAt = project.createdAt.getOrElse(Timestamp.from(Instant.now())).getTime
+      Logger.info(s"Found project: ${project.ownerName}/${project.slug}")
```
Ewoutvans: why info and not debug?

build.sbt
```diff
@@ -25,6 +25,7 @@ libraryDependencies ++= Seq(
   "org.bouncycastle"      %   "bcpg-jdk15on"            %   "1.60",
   "javax.mail"            %   "mail"                    %   "1.4.7",
   "com.beachape"          %%  "enumeratum"              %   "1.5.13",
+  "com.chuusai"           %% "shapeless"                %   "2.3.3",
```
Ewoutvans: not aligned with rest of the dependencies 😃

Ewoutvans (approved):

app/ore/project/ProjectTask.scala
Katrix: Ops, I'll fix that

app/views/projects/versions/log.scala.html
```diff
@@ -42,7 +42,7 @@ <h4 class="panel-title pull-left">@messages("version.log.visibility.title")</h4>
 							@visibilityChanges.reverse.map { case (entry, createdBy) =>
 							<tr>
 								<td>@VisibilityTypes.withId(entry.visibility)</td>
-								<td>@prettifyDateAndTime(entry.createdAt.getOrElse(Timestamp.from(Instant.now())))</td>
+								<td>@prettifyDateAndTime(entry.createdAt.value))))</td>
```
Ewoutvans: to many braces

Ewoutvans (approved):

Ewoutvans (approved):

Ewoutvans (approved):

Ewoutvans (approved):

Ewoutvans (approved):

felixoi (approved):

PR #570, opened 2018-07-15
Explain in the delete modal that we reserve the unique id

Ewoutvans:
Before:
schermafbeelding 2018-07-15 om 11 53 59

Updated:
schermafbeelding 2018-07-15 om 17 14 45

Old version in PR:
schermafbeelding 2018-07-15 om 12 01 14

### DIFF
--- a/app/views/projects/settings.scala.html
+++ b/app/views/projects/settings.scala.html
@@ -262,7 +262,12 @@
                 @form(action = projectRoutes.softDelete(p.project.ownerName, p.project.slug)) {
                 <div class="modal-body">
                     @messages("project.delete.info")
+                    <br>
                     <textarea name="comment" class="textarea-delete-comment form-control" rows="3"></textarea>
+                    <br>
+                    <div class="alert alert-warning">
+                        @messages("project.delete.info.uniqueid", p.project.name)
+                    </div>
                 </div>
                 <div class="modal-footer">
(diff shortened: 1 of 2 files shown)


dualspiral:
I don't think this is that clear - reserving the project ID might mean that the author has the sole rights to it, not the case. We must be explicit. I also think it should be in a stronger colour, yellow or red to indicate danger.

"WARNING: You will not be able to use the plugin ID "{pluginid}" in the future if you continue. If you are deleting your project to recreate it, please do not delete your project and contact the Ore staff for help."

app/views/projects/settings.scala.html
```diff
@@ -262,7 +262,12 @@ <h4 class="modal-title" id="label-delete">@messages("project.delete.title")</h4>
                 @form(action = projectRoutes.softDelete(p.project.ownerName, p.project.slug)) {
                 <div class="modal-body">
                     @messages("project.delete.info")
+                    <br>
                     <textarea name="comment" class="textarea-delete-comment form-control" rows="3"></textarea>
+                    <br>
+                    <div class="alert alert-warning">
+                        @Html(messages("project.delete.info.uniqueid", p.project.name))
```
Katrix: Is this Html call safe?

Ewoutvans: I believe so

PR #572, opened 2018-07-15
don't require all groups to exist in RoleTypes

kashike:


app/ore/permission/role/RoleTypes.scala
```diff
@@ -69,11 +69,7 @@ object RoleTypes extends Enumeration {
     * @param internalName Internal name
     * @return UserRole with specified internal name
     */
-  def withInternalName(internalName: String): RoleType = this.values.find(_.internalName == internalName).getOrElse {
-    // Throw an exception instead of returning an Option to match Enumeration
-    // behavior
-    throw new NoSuchElementException
-  }
+  def withInternalName(internalName: String): Option[Value] = this.values.find(_.internalName == internalName)
```
Katrix: Keep the type as RoleType. Not sure if having it be Value can screw over other stuff.

Ewoutvans (approved):
Tested; Works

app/controllers/ApiController.scala
```diff
@@ -309,7 +309,11 @@ final class ApiController @Inject()(api: OreRestfulApi,
               if (groups.trim == "")
                 Set.empty
               else
-                groups.split(",").map(group => RoleTypes.withInternalName(group)).toSet[RoleType]
+                groups.split(",")
+                  .map(group => RoleTypes.withInternalName(group))
+                  .filter(group => group.nonEmpty)
```
Katrix: Use flatMap here instead of filter + map, or use collect

PR #575, opened 2018-07-18
Reorganizing Permissions for Moderators

phase:
This should give everyone the proper permissions.
Moderators uploading versions was the main issue.

### DIFF
--- a/app/ore/permission/role/RoleTypes.scala
+++ b/app/ore/permission/role/RoleTypes.scala
@@ -15,19 +15,19 @@ object RoleTypes extends Enumeration {
 
   // Global
 
-  val Admin           = new  RoleType( 0, "Ore_Admin",         61, classOf[GlobalRole], Absolute, "Ore Admin",          Red)
-  val Mod             = new  RoleType( 1, "Ore_Moderator",     62, classOf[GlobalRole], Lifted,   "Ore Moderator",      Aqua)
-  val SpongeLeader    = new  RoleType( 2, "Sponge_Leader",     44, classOf[GlobalRole], Default,  "Sponge Leader",      Amber)
-  val TeamLeader      = new  RoleType( 3, "Team_Leader",       58, classOf[GlobalRole], Default,  "Team Leader",        Amber)
-  val CommunityLeader = new  RoleType( 4, "Community_Leader",  59, classOf[GlobalRole], Default,  "Community Leader",   Amber)
-  val Staff           = new  RoleType( 5, "Sponge_Staff",      3,  classOf[GlobalRole], Default,  "Sponge Staff",       Amber)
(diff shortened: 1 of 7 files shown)


app/ore/permission/role/RoleTypes.scala
```diff
@@ -38,18 +38,18 @@ object RoleTypes extends Enumeration {
   // Project
 
   val ProjectOwner   = new RoleType(17, "Project_Owner",    -1, classOf[ProjectRole], Absolute, "Owner",    Transparent, isAssignable = false)
-  val ProjectDev     = new RoleType(18, "Project_Developer",-2, classOf[ProjectRole], Standard, "Developer",Transparent)
+  val ProjectDev     = new RoleType(18, "Project_Developer",-2, classOf[ProjectRole], Lifted,   "Developer",Transparent)
```
dualspiral: I'm concerned about this, because we're potentially giving admin permissions to those marked as developers on project teams when this wasn't the case before, **particularly user admin** - this in itself may be a problem in some teams. It looks to me (and tell me if I'm wrong) that a developer could invite a person onto the team when they couldn't before and give them developer access. Worse, this wouldn't be a visible change and would have to be noticed, if it's noticed by the wrong person, this could be a serious problem.

I know there are other changes coming to Ore that should help fix this permissions problem long term, but right now, this isn't a good solution. It trades one problem with another.

Is there some way we can mitigate this?

phase: We could make `EditSettings` an `Absolute` permission.

dualspiral: I'm wary of granting **any** permissions like this. Developers in projects should not gain any already exisiting permissions **ever**. I would argue that this doing so is a security problem because we are changing the very definition of the role of developer by giving these admin roles.

The Developer role, where given within a project or org, has a set of permissions that the project owner was comfortable giving them. `UserAdmin`, which I assume is inviting users to the group, is not one of them. If they suddenly get that, or changing API keys, that's a breach of trust on our end that we're going to protect someone's project using appropriate controls.

Also, how will moving `EditSettings` to `Absolute` change things within the rest of Ore? Would that impede any other high-level moderators work?

My feeling is that moving developer from `moderator` to `lifted` is probably not a good move, though I know the problem is they then cannot upload versions, which is the motivation for this. Can we add a role type in the middle, say "publisher", that contains the upload permission?

dualspiral:
Yep, that looks reasonable now, thanks for hearing me out!

dualspiral (approved):
I don't know Scala that well (someone should gloss their eyes over it from that side of things). However, it looks OK for the issue it solves.

Ewoutvans (approved):

PR #610, opened 2018-08-07
Add page filter

### LINKED ISSUE #602: User Action Log information improvements
Add links to visit the actual:
  - [x] Project
  - [x] Page
  - [x] Version

Show project name, page name (versions appear to be shown already) in the log entry:
  - [x] Show project name
  - [x] Show page name

When filtering:
  - [ ] Show project name and provide link to project
  - [ ] Show page name and provide link to page
  - [ ] Show version name and provide link to version

Ewoutvans:
schermafbeelding 2018-08-07 om 15 25 11

 fixes #602

### DIFF
--- a/conf/evolutions/default/97.sql
+++ b/conf/evolutions/default/97.sql
@@ -1,6 +1,6 @@
 # --- !Ups
 
-UPDATE logged_actions SET action_context_id = -1 where action_context = 2;
+DELETE FROM logged_actions WHERE action_context = 2;
 
 DROP VIEW v_logged_actions;
 
@@ -29,7 +29,7 @@ SELECT
    pp.slug            as pp_slug,
    CASE
        WHEN (a.action_context = 0) THEN a.action_context_id         -- Project
-       WHEN (a.action_context = 1) THEN coalesce(pv.project_id, -1) -- Version
+       WHEN (a.action_context = 1) THEN COALESCE(pv.project_id, -1) -- Version
        WHEN (a.action_context = 2) THEN a.action_context_id         -- ProjectPage
        ELSE -1 -- Return -1 to allow filtering
    END as filter_project,
@@ -45,21 +45,13 @@ FROM logged_actions a
 LEFT OUTER JOIN users u ON a.user_id = u.id
 LEFT OUTER JOIN projects p ON
    CASE
-       WHEN a.action_context IN (0) AND a.action_context_id = p.id THEN 1 -- Join on action
-       WHEN a.action_context IN (1) AND (SELECT project_id FROM project_versions pvin WHERE pvin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Version
-       WHEN a.action_context IN (2) AND (SELECT project_id FROM project_pages ppin WHERE ppin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Page
+       WHEN a.action_context = 0 AND a.action_context_id = p.id THEN 1 -- Join on action
+       WHEN a.action_context = 1 AND (SELECT project_id FROM project_versions pvin WHERE pvin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Version
+       WHEN a.action_context = 2 AND (SELECT project_id FROM project_pages ppin WHERE ppin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Page
        ELSE 0
    END = 1
-LEFT OUTER JOIN project_versions pv ON
-   CASE
-       WHEN a.action_context IN (1) AND a.action_context_id = pv.id THEN 1
-       ELSE 0
-   END = 1
-LEFT OUTER JOIN project_pages pp ON
-   CASE
-       WHEN a.action_context IN (2) AND a.action_context_id = pp.id THEN 1
-       ELSE 0
-   END = 1
+LEFT OUTER JOIN project_versions pv ON (a.action_context = 1 AND a.action_context_id = pv.id)
+LEFT OUTER JOIN project_pages pp ON (a.action_context = 2 AND a.action_context_id = pp.id)
 ;
 # --- !Downs
 
--- a/conf/evolutions/default/97.sql
+++ b/conf/evolutions/default/97.sql
@@ -1,0 +1,113 @@
+# --- !Ups
+
+UPDATE logged_actions SET action_context_id = -1 where action_context = 2;
+
+DROP VIEW v_logged_actions;
+
+CREATE VIEW v_logged_actions
+   AS
+SELECT
+   a.id,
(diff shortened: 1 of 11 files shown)


phase (approved):

conf/evolutions/default/97.sql
```diff
@@ -0,0 +1,113 @@
...
+   END as filter_version,
+   CASE
+       WHEN (a.action_context = 2) THEN COALESCE(pp.id, -1)
+       ELSE -1
+   END as filter_page
+FROM logged_actions a
+LEFT OUTER JOIN users u ON a.user_id = u.id
+LEFT OUTER JOIN projects p ON
+   CASE
+       WHEN a.action_context IN (0) AND a.action_context_id = p.id THEN 1 -- Join on action
```
dualspiral: Something that particularly jumps out at me is that you're using `IN` for a single entry comparison. It might be that postgres will optimise this away, but you may get slightly better results putting `a.action_context=0`, rather than `IN`.

conf/evolutions/default/97.sql
```diff
@@ -0,0 +1,113 @@
...
+       WHEN a.action_context IN (1) AND (SELECT project_id FROM project_versions pvin WHERE pvin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Version
+       WHEN a.action_context IN (2) AND (SELECT project_id FROM project_pages ppin WHERE ppin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Page
+       ELSE 0
+   END = 1
+LEFT OUTER JOIN project_versions pv ON
```
dualspiral: Check this, but instead of the `CASE`, maybe

```sql
LEFT OUTER JOIN project_versions pv ON (a.action_context=1 AND a.action_context_id = pv.id)
```

That way, you're removing the need for an extra comparison, and making it clearer to the optimiser what you're asking for.

In the case of multiple `CASE`s, you could use a combination of `AND` and `OR`, but it'll be less readable and it may not really be better.

It could be that postgres is smart enough to optimise this well as is - explain query plans could help here.

conf/evolutions/default/97.sql
```diff
@@ -0,0 +1,113 @@
...
+   p.owner_name       as p_owner_name,
+   pv.id              as pv_id,
+   pv.version_string  as pv_version_string,
+   pp.id              as pp_id,
+   pp.name            as pp_name,
+   pp.slug            as pp_slug,
+   CASE
+       WHEN (a.action_context = 0) THEN a.action_context_id         -- Project
+       WHEN (a.action_context = 1) THEN coalesce(pv.project_id, -1) -- Version
+       WHEN (a.action_context = 2) THEN a.action_context_id         -- ProjectPage
```
dualspiral: I couldn't help but notice your first line made anything with `action_context=2` be given an `action_context_id=-1`, so this is always going to return a constant, indeed, the same constant in the `ELSE` block. Perhaps comment out/remove this line? It's a part of the query that you don't need to run.

conf/evolutions/default/97.sql
```diff
@@ -0,0 +1,113 @@
...
+   p.slug             as p_slug,
+   p.owner_name       as p_owner_name,
+   pv.id              as pv_id,
+   pv.version_string  as pv_version_string,
+   pp.id              as pp_id,
+   pp.name            as pp_name,
+   pp.slug            as pp_slug,
+   CASE
+       WHEN (a.action_context = 0) THEN a.action_context_id         -- Project
+       WHEN (a.action_context = 1) THEN coalesce(pv.project_id, -1) -- Version
```
dualspiral: coalesce -> COALESCE for consistency

conf/evolutions/default/97.sql
```diff
@@ -0,0 +1,113 @@
...
+       WHEN a.action_context IN (1) AND (SELECT project_id FROM project_versions pvin WHERE pvin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Version
+       WHEN a.action_context IN (2) AND (SELECT project_id FROM project_pages ppin WHERE ppin.id = a.action_context_id) = p.id THEN 1 -- Query for projectId from Page
+       ELSE 0
+   END = 1
+LEFT OUTER JOIN project_versions pv ON
+   CASE
+       WHEN a.action_context IN (1) AND a.action_context_id = pv.id THEN 1
+       ELSE 0
+   END = 1
+LEFT OUTER JOIN project_pages pp ON
```
dualspiral: Similar comment to above.

dualspiral (commented):
These are all merely suggestions, but as you have a complex view, I suspect this might get quite slow as time progresses.

It looks OK to go, but consider making some changes and running explain plans on them to get a feel for what might be a bottleneck, what might work better, and how you may improve it either here or elsewhere, now or in the future.

dualspiral:
I still have a query about the SQL (har har), but before I open my mouth I figure I'd be better asking a question first to help me understand what's going on here, and Ore in general.

In the first line of your Ups, you have the line:

```sql
DELETE FROM logged_actions WHERE action_context = 2;
```

This looks like it's either deleting items from the log, or you're removing this action type from the log completely. Without knowing what action_context 2 is meant to be, I'm not sure exactly what's going on.

* What is `action_context` 2 (what does it represent)?
* What are you deleting? (this is a sanity check question!)
* Will there be any entries with `action_context = 2` once this goes live and this has been running for a while?

The first question is informational, the second and third are relavent to what I've spotted. If we will _never_ have `action_context=2`, then there are a few lines in the view that should be removed (including a join), as they depend on this condition which will always be false.

If the `action_context` _can_ be `2`, why delete entries when it was previously an `UPDATE`?

I eventually hope to get an Ore dev setup locally to inspect the databases a little more so I can avoid asking questions about this if a database question comes up, but without that knowledge right now, I just want to make sure that the right thing is being done!

Ewoutvans:
So action_context 2 is for tracking project_pages (edits / creating / deleting) but in the original design we saved the action_context_id as the project_id and not the page_id which was fine and didn't cause any issues. But in issue #602 the request is to filter on pages and link to the pages but because we didn't save the page_id we have no way of knowing what the page was.
This PR fixes that action_context 2 now tracks the page_id and not the project_id. because of this behaviour change I have decided it is better to delete the old ones (originally I was going to updated it to -1 but this gave more problems because the data became unreliable. My reasoning for the deleting is that Ore is still not in its 'review' stage and this loss will have non to little impact).

So when this script runs it will delete the old ones and introduce the new correct behaviour.

dualspiral:
Right, got it! In that case, no qualms, as far as the SQL goes, it looks good (assuming it’s tested).

dualspiral (approved):
Approval for SQL only (as that’s all I looked at)

Issue #527, opened 2018-04-15
BungeeCord Plugin support


phit:
Came up when talking about [SpongePls](https://forums.spongepowered.org/t/spongepls-bungeecord-forge-support/9891).
Ore support to parse Bungee plugins would be a nice addition in the future.
They should be fairly easy to detect as they all have a text file named `bungee.yml` in the jar root. 

e.g.
```
name: 'SpongePls'
main: 'co.ryred.spongepls.SpongePlsPlugin'
author: 'Cory Redmond'
description: 'SpongePls allows IPForwarding to sponge servers.'
version: '1.1-SNAPSHOT-b#35'
```

simon816:
To me this seems out of scope for Ore.
If you're looking for bungeecord plugins then you visit the bungeecord plugin list (https://www.spigotmc.org/resources/categories/bungee-proxy.3/)
Also I'm not sure we would want this to be mixed with other plugins/mods on Ore, e.g. what is the plugin ID? What happens when there's a bungeecord plugin of the same name as a sponge plugin?
Should they conform to the guidelines, would the review process be the same?

phase:
I think for now we want to focus our attention on mods for our API, Sponge, and later down the road, if there are a large number of people wanting to host on our platform, we can add support for other types of mods. This and #510 can be scheduled for a v2.2.0 or v3.0 release (maybe v3.0 can focus on specifically hosting other mods?).

progwml6:
Forge mod support can come with plugins, they are handled nearly identically

ryantheleach:
Should they conform to the guidelines, would the review process be the same?

Yes. Yes.

>What happens when there's a bungeecord plugin of the same name as a sponge plugin?

If they are related, then the same way that we currently bundle multiple artifacts under one brand/name?

If unrelated, it would have to follow the same way we already defined how to handle name collisions.

I don't see it as any different to a Forge mod or Sponge plugin, People are interested in having Sponge ecosystem specific proxy plugins that are compatible with Bungee, They need a place to promote them. The Spigot/Bungee Community is largely hostile against Sponge users, so why shouldn't we allow the **authors** of Bungee plugins that wish to host on our services the ability to do so, if they work well or complement our platform?

It's a no brainer to me to allow it, as long as it follows all previously set out guidelines.

phase:
This might come sooner than expected. Support for this would not be difficult to add, and I am absolutely open to having other platforms hosted on Ore.

ryantheleach:
[...]

Issue #510, opened 2018-03-17
Forge Mod Support


phase:
Forge tags exist for Sponge plugins that have Forge as a dependency, but standalone Forge mod support is limited.

* [x] Forge Tag
* [ ] Get Minecraft Version and add a Minecraft Version tag

phase:
Some plugins like to override the generation of the `mcmod.info` file with their own. Currently, mods and plugins like this won't appear with any tags. I think they should be rejected, as we're unable to determine what platform the mod is for.

progwml6:
the other option is to use ASM to find the forge/sponge mod annotations on upload, however ... i'd rather avoid doing this, as i don't want to be scanning for this if 1.13 fml changes break the current setup

PR #646, opened 2018-08-20
Rewrite Metadata Ripper to support multiple types of files, and add Mixin Tag

### LINKED ISSUE #247: Add Mixin tag onto Projects
After reading through https://github.com/SpongePowered/SpongeVanilla/issues/271 I thought I would propose adding a Mixin tag to the project list.

(Example applied to the project "Nucleus Mixins" on Ore)
example

This could be a selective or automatic process by looking at the jars manifest for the relevant Mixin entries on the upload of the plugin.

This would allow server administrators to become aware that the project they're downloading contains a Mixin (reason is defined in the linked issue in SpongeVanilla), though the example I used is very obvious, other cases may not be.

Having this without some kind of explanation is useless, thus this information should be clearly provided in the docs or anywhere else that best fits so that server administrators can clearly understand.

phase:
The old system was completely reliant on Sponge's `mcmod.info` parser.
We can't rely on Sponge's implementation because it'll probably be removed once
1.13 comes out. It has been replaced with a more generic metadata system that can
work for any type of file, as long as there is a parser for it.

`PluginFileData` contains a list of `DataValue`s that have different meta about
the project. The majority of it is platform agnostic, such as plugin id &
version, but some things are super specific like `MixinConfigs` being present in
the `MANIFEST.MF`. The system supports all kinds of data, and migrating to the
Forge 1.13 TOML format will take plugging in a TOML parser. This opens the door
to other parsing other platform metadata files like BungeeCoord's `bungee.yml`.

Dependency parsing is a little different now since we're not relying on Sponge's 
parser. It's essentially the same, except dependencies are parsed into the
`PluginFileData` and then put into the `Version` when it is constructed.
This is important to note for future platforms that don't have a `requiredMods`
section. All you would need to do is add a dependency list within your file 
handler. So at the top of the `bungee.yml` handler, slap a `DependencyDataValue`
containing the platform. The platform will get parsed into a tag later in the
process.

This PR relies on the `feature/sf-tags` branch. If that is merged in first then
we can just change the base branch of this PR.

Closes #247 
References #527 
References #510

### DIFF
--- a/app/ore/project/factory/ProjectFactory.scala
+++ b/app/ore/project/factory/ProjectFactory.scala
@@ -364,7 +364,7 @@ trait ProjectFactory {
         )
         this.service.access[Version](classOf[Version]).add(newVersion)
       }
-      tags <- addDependencyTags(newVersion)
+      tags <- addTags(pending, newVersion)
     } yield {
       // Notify watchers
       this.actorSystem.scheduler.scheduleOnce(Duration.Zero, NotifyWatchersTask(newVersion, project))
@@ -381,9 +381,30 @@ trait ProjectFactory {
     }
   }
 
+  private def addTags(pendingVersion: PendingVersion, newVersion: Version)(implicit ec: ExecutionContext): Future[Seq[ProjectTag]] = {
+    for {
+      metadataTags <- addMetadataTags(pendingVersion.plugin.data, newVersion)
+      dependencyTags <- addDependencyTags(newVersion)
+    } yield {
+      metadataTags ++ dependencyTags
+    }
+  }
+
+  private def addMetadataTags(pluginFileData: Option[PluginFileData], version: Version)(implicit ec: ExecutionContext): Future[Seq[ProjectTag]] = {
+    for {
+      tags <- Future.sequence(pluginFileData.map(_.getGhostTags.map(_.getFilledTag(service))).toList.flatten)
+    } yield {
+      tags.map { tag =>
+        tag.addVersionId(version.id.get)
+        version.addTag(tag)
+        tag
+      }
+    }
+  }
+
   private def addDependencyTags(version: Version)(implicit ec: ExecutionContext): Future[Seq[ProjectTag]] = {
     for {
-      tags <- Future.sequence(Platforms.getPlatformGhostTags(version.dependencies).map(tag => tag.getFilledTag(service)))
+      tags <- Future.sequence(Platforms.getPlatformGhostTags(version.dependencies).map(_.getFilledTag(service)))
     } yield {
       tags.map { tag =>
         tag.addVersionId(version.id.get)
--- a/app/ore/project/factory/ProjectFactory.scala
+++ b/app/ore/project/factory/ProjectFactory.scala
@@ -8,36 +8,32 @@ import com.google.common.base.Preconditions._
 import db.ModelService
 import db.impl.OrePostgresDriver.api._
 import db.impl.access.{ProjectBase, UserBase}
+import db.impl.{ProjectMembersTable, ProjectRoleTable}
 import discourse.OreDiscourseApi
 import javax.inject.Inject
-import models.project.TagColors.TagColor
 import models.project._
 import models.user.role.ProjectRole
 import models.user.{Notification, User}
 import ore.Colors.Color
-import ore.{OreConfig, OreEnv, Platforms}
 import ore.permission.role.RoleTypes
-import ore.project.Dependency
-import ore.project.{NotifyWatchersTask, ProjectMember}
 import ore.project.factory.TagAlias.ProjectTag
-import ore.project.io.{InvalidPluginFileException, PluginFile, PluginUpload, ProjectFiles}
+import ore.project.io._
+import ore.project.{NotifyWatchersTask, ProjectMember}
+import ore.user.MembershipDossier
 import ore.user.notification.NotificationTypes
-import org.spongepowered.plugin.meta.PluginMetadata
+import ore.{OreConfig, OreEnv, Platforms}
 import play.api.cache.SyncCacheApi
 import play.api.i18n.Messages
 import security.pgp.PGPVerifier
 import util.StringUtils._
-import util.functional.{EitherT, OptionT}
+import util.functional.EitherT
 import util.instances.future._
 import util.syntax._
 
-import scala.collection.JavaConverters._
-import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration.Duration
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 import scala.util.matching.Regex
-import db.impl.{ProjectMembersTable, ProjectRoleTable}
-import ore.user.MembershipDossier
 
 /**
   * Manages the project and version creation pipeline.
@@ -68,7 +64,7 @@ trait ProjectFactory {
     * @param owner      Upload owner
     * @return Loaded PluginFile
     */
-  def processPluginUpload(uploadData: PluginUpload, owner: User)(implicit messages: Messages): PluginFile = {
+  def processPluginUpload(uploadData: PluginUpload, owner: User)(implicit messages: Messages): Either[PluginFile, String] = {
     val pluginFileName = uploadData.pluginFileName
     var signatureFileName = uploadData.signatureFileName
 
@@ -103,32 +99,45 @@ trait ProjectFactory {
 
     // create and load a new PluginFile instance for further processing
     val plugin = new PluginFile(pluginPath, sigPath, owner)
-    plugin.loadMeta()
-    plugin
+    val result = plugin.loadMeta()
+    result match {
+      case Left(_) => Left(plugin)
+      case Right(errorMessage) => Right(errorMessage)
+    }
   }
 
   def processSubsequentPluginUpload(uploadData: PluginUpload,
                                     owner: User,
                                     project: Project)(implicit ec: ExecutionContext, messages: Messages): EitherT[Future, String, PendingVersion] = {
-    val plugin = this.processPluginUpload(uploadData, owner)
-    if (!plugin.meta.get.getId.equals(project.pluginId))
-      EitherT.leftT("error.version.invalidPluginId")
-    else {
-      EitherT(
-        for {
-          (channels, settings) <- (project.channels.all, project.settings).parTupled
-          version = this.startVersion(plugin, project, settings, channels.head.name)
-          modelExists <- version.underlying.exists
-        } yield {
-          if (modelExists && this.config.projects.get[Boolean]("file-validate"))
-            Left("error.version.duplicate")
-          else {
-            version.cache()
-            Right(version)
+    this.processPluginUpload(uploadData, owner) match {
+      case Left(plugin) => if (!plugin.data.flatMap(_.getId).contains(project.pluginId))
+        EitherT.leftT("error.version.invalidPluginId")
+      else {
+        EitherT(
+          for {
+            (channels, settings) <- (project.channels.all, project.settings).parTupled
+            version = this.startVersion(plugin, project, settings, channels.head.name)
+            modelExists <- version match {
+              case Left(v) => v.underlying.exists
+              case Right(_) => Future.successful(false)
+            }
+          } yield {
+            version match {
+              case Left(v) => if (modelExists && this.config.projects.get[Boolean]("file-validate"))
+                Left("error.version.duplicate")
+              else {
+                v.cache()
+                Right(v)
+              }
+              case Right(m) => Left(m)
+            }
+
           }
-        }
-      )
+        )
+      }
+      case Right(errorMessage) => new EitherT[Future, String, PendingVersion](Future.successful(Left[String, PendingVersion](errorMessage)))
     }
+
   }
 
   /**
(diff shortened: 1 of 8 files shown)


app/ore/project/factory/ProjectFactory.scala
```diff
@@ -68,7 +64,7 @@ trait ProjectFactory {
     * @param owner      Upload owner
     * @return Loaded PluginFile
     */
-  def processPluginUpload(uploadData: PluginUpload, owner: User)(implicit messages: Messages): PluginFile = {
+  def processPluginUpload(uploadData: PluginUpload, owner: User)(implicit messages: Messages): Either[PluginFile, String] = {
```
Katrix: Switch these, right is for "correct", left is for errors

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -103,32 +99,45 @@ trait ProjectFactory {
...
-          version = this.startVersion(plugin, project, settings, channels.head.name)
-          modelExists <- version.underlying.exists
-        } yield {
-          if (modelExists && this.config.projects.get[Boolean]("file-validate"))
-            Left("error.version.duplicate")
-          else {
-            version.cache()
-            Right(version)
+    this.processPluginUpload(uploadData, owner) match {
+      case Left(plugin) => if (!plugin.data.flatMap(_.getId).contains(project.pluginId))
```
Katrix: Place this if in the match clause, and make a new clause if the test doesn't hold

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -103,32 +99,45 @@ trait ProjectFactory {
...
+          for {
+            (channels, settings) <- (project.channels.all, project.settings).parTupled
+            version = this.startVersion(plugin, project, settings, channels.head.name)
+            modelExists <- version match {
+              case Left(v) => v.underlying.exists
+              case Right(_) => Future.successful(false)
+            }
+          } yield {
+            version match {
+              case Left(v) => if (modelExists && this.config.projects.get[Boolean]("file-validate"))
```
Katrix: Place this if in the match clause, and make a new clause if the test doesn't hold

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -103,32 +99,45 @@ trait ProjectFactory {
...
+              }
+              case Right(m) => Left(m)
+            }
+
           }
-        }
-      )
+        )
+      }
+      case Right(errorMessage) => new EitherT[Future, String, PendingVersion](Future.successful(Left[String, PendingVersion](errorMessage)))
```
Katrix: Use EitherT.<side>T\[Future, <other side's type>\](errorMessage) instead

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -188,19 +197,17 @@ trait ProjectFactory {
     * @param project Parent project
     * @return PendingVersion instance
     */
-  def startVersion(plugin: PluginFile, project: Project, settings: ProjectSettings, channelName: String): PendingVersion = {
+  def startVersion(plugin: PluginFile, project: Project, settings: ProjectSettings, channelName: String): Either[PendingVersion, String] = {
```
Katrix: Switch these, right is for "correct", left is for errors

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -374,9 +381,30 @@ trait ProjectFactory {
     }
   }
 
+  private def addTags(pendingVersion: PendingVersion, newVersion: Version)(implicit ec: ExecutionContext): Future[Seq[ProjectTag]] = {
+    for {
+      metadataTags <- addMetadataTags(pendingVersion.plugin.data, newVersion)
```
Katrix: parTupled?

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -374,9 +381,30 @@ trait ProjectFactory {
...
+      dependencyTags <- addDependencyTags(newVersion)
+    } yield {
+      metadataTags ++ dependencyTags
+    }
+  }
+
+  private def addMetadataTags(pluginFileData: Option[PluginFileData], version: Version)(implicit ec: ExecutionContext): Future[Seq[ProjectTag]] = {
+    for {
```
Katrix: Is a for comprehension needed here?

app/ore/project/io/PluginFile.scala
```diff
@@ -78,58 +75,64 @@ class PluginFile(private var _path: Path, val signaturePath: Path, val user: Use
...
       // Find plugin meta file
       var entry: JarEntry = null
-      var metaFound: Boolean = false
-      breakable {
-        while ({ entry = jarIn.getNextJarEntry; entry } != null) {
-          if (entry.getName.equals(MetaFileName)) {
-            metaFound = true
-            break
-          }
+      while ( {
```
Katrix: Can we rewrite this as tailrec?

app/ore/project/io/PluginFile.scala
```diff
@@ -78,58 +75,64 @@ class PluginFile(private var _path: Path, val signaturePath: Path, val user: Use
...
+        entry = jarIn.getNextJarEntry
+        entry
+      } != null) {
+        if (fileNames.contains(entry.getName)) {
+          data = data ::: PluginFileData.getData(entry.getName, new BufferedReader(new InputStreamReader(jarIn)))
```
Katrix: ::: on a list in a loop is bad as it needs to go over the entire list. Can we use another data structure, or refactor this to be able to use ::?

app/ore/project/io/PluginFile.scala
```diff
@@ -78,58 +75,64 @@ class PluginFile(private var _path: Path, val signaturePath: Path, val user: Use
     *
     * TODO: More validation on PluginMetadata results (null checks, etc)
     *
-    * @return Result of parse
+    * @return Plugin metadata or an error message
     */
   @throws[InvalidPluginFileException]
-  def loadMeta()(implicit messages: Messages): PluginMetadata = {
+  def loadMeta()(implicit messages: Messages): Either[PluginFileData, String] = {
```
Katrix: Switch these, right is for "correct", left is for errors

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+  * The metadata within a [[PluginFile]]
+  *
+  * @author phase
+  * @param data the data within a [[PluginFile]]
+  */
+class PluginFileData(var data: List[DataValue[_]]) {
+
+  data = data.groupBy(_.key).flatMap { case (key, value) =>
```
Katrix: Can we make a factory that does this instead and get away with a val instead of var?

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+
+import scala.collection.mutable.ListBuffer
+
+/**
+  * The metadata within a [[PluginFile]]
+  *
+  * @author phase
+  * @param data the data within a [[PluginFile]]
+  */
+class PluginFileData(var data: List[DataValue[_]]) {
```
Katrix: Can we use the scala naming convention for the accessors in this class?

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+}
+
+/**
+  * A data element in a data file
+  *
+  * @param key   the key for the value
+  * @param value the value extracted from the file
+  * @tparam T the type of the value
+  */
+sealed class DataValue[T](val key: String, val value: T)
```
Katrix: Make this a sealed trait instead

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+  * A data element that is a String, such as the plugin id or version
+  *
+  * @param value the value extracted from the file
+  */
+case class StringDataValue(override val key: String, override val value: String)
```
Katrix: No need for override val here

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+  extends DataValue[String](key, value)
+
+/**
+  * A data element that is a list of strings, such as an authors list
+  *
+  * @param value the value extracted from the file
+  */
+case class StringListValue(override val key: String, override val value: List[String])
```
Katrix: No need for override val here

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+  extends DataValue[List[String]](key, value)
+
+/**
+  * A data element that is a list of [[Dependency]]
+  *
+  * @param value the value extracted from the file
+  */
+case class DependencyDataValue(override val key: String, override val value: List[Dependency])
```
Katrix: No need for override val here

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+  extends DataValue[List[Dependency]](key, value)
+
+sealed abstract case class FileType(fileName: String) {
+  def getData(bufferedReader: BufferedReader): List[DataValue[_]]
+}
+
+object ModInfo extends FileType("mcmod.info") {
+  override def getData(bufferedReader: BufferedReader): List[DataValue[_]] = {
+
+    val dataValues = new ListBuffer[DataValue[_]]
```
Katrix: ListBuffer has bad performance, use ArrayBuffer

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+          for (j <- 0 until authorsJson.size()) {
+            val element = authorsJson.get(j)
+            authors += element.getAsString
+          }
+          dataValues += StringListValue("authors", authors.toList)
+        }
+
+        // collect dependencies
+
+        val dependencies = new ListBuffer[Dependency]
```
Katrix: ListBuffer has bad performance, use ArrayBuffer

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,216 @@
...
+
+    dataValues.toList
+  }
+}
+
+object Manifest extends FileType("META-INF/MANIFEST.MF") {
+  override def getData(bufferedReader: BufferedReader): List[DataValue[_]] = {
+    val dataValues = new ListBuffer[DataValue[_]]
+
+    val lines = Stream.continually(bufferedReader.readLine()).takeWhile(_ != null)
```
Katrix: Stream is considered one of the "bad" things in the scala library. Use NIO instead.

Katrix (changes requested):
Frustrated?

stephan-gh:
> The old system was completely reliant on Sponge's `mcmod.info` parser.
We can't rely on Sponge's implementation because it'll probably be removed once 1.13 comes out. It has been replaced with a more generic metadata system that can work for any type of file, as long as there is a parser for it.

This is exactly what `plugin-meta` is: an abstraction on top of `mcmod.info` or other metadata files. Granted, it is somewhat specific to Sponge's needs because it primarily exposes data that is relevant for Sponge plugins. However, it does have an additional `extensions` object, where additional metadata can be stored and accessed. (We used to use this for some Sponge specific extensions but not anymore currently...)

It has never exposed `mcmod.info` as-is, because that format is rather weird to work with. For example dependencies are exposed completely differently.

I see no reason to remove it in 1.13: Many of our tools (SpongeAPI, SpongeVanilla, SpongeGradle, ...) depend on `plugin-meta`. It can easily be extended with a TOML parser/emitter so we only need to do minimal changes everywhere else (basically only changing file names and swapping the parser/emitter...).

So in my opinion, creating another parser in Ore is unneeded duplication. `plugin-meta` should be able to handle most metadata formats to some extent, assuming that someone writes a parser for them. Could you clarify what exactly is missing in `plugin-meta`?

Katrix:
From my POV, as long as plugin-meta continues to support the old mcmod.info files, we can keep it. However if it drops that support, then I don't really want to deal with jar hell to use both a version of it that support mcmod.info, and another one that supports toml.

stephan-gh:
@Katrix I see no reason to drop `mcmod.info` support in plugin-meta. The parser etc is already separated from the main `PluginMetadata` classes, so we can just add TOML support additionally.

phase:
Well that's great news! I'll switch it back to using `plugin-meta`, and then we can add on Forge TOML support once it's added.

app/ore/project/factory/ProjectFactory.scala
```diff
@@ -103,32 +99,45 @@ trait ProjectFactory {
...
-          version = this.startVersion(plugin, project, settings, channels.head.name)
-          modelExists <- version.underlying.exists
-        } yield {
-          if (modelExists && this.config.projects.get[Boolean]("file-validate"))
-            Left("error.version.duplicate")
-          else {
-            version.cache()
-            Right(version)
+    this.processPluginUpload(uploadData, owner) match {
+      case Right(plugin) => if (!plugin.data.flatMap(_.getId).contains(project.pluginId))
```
Katrix: Turn that if test into a guard in the match expression

app/ore/project/io/PluginFileData.scala
```diff
@@ -0,0 +1,184 @@
...
+
+import scala.collection.mutable.ArrayBuffer
+
+/**
+  * The metadata within a [[PluginFile]]
+  *
+  * @author phase
+  * @param data the data within a [[PluginFile]]
+  */
+class PluginFileData(data: Seq[DataValue[_]]) {
```
Katrix, line 18: Can we use scala naming conventions for this class? `x` instead of `getX`

phase: :relaxed: my Java is showing

Katrix (approved):

PR #648, opened 2018-08-22
Undelete version & hard delete version

### LINKED ISSUE #479: Moderator access to deleted versions

Ewoutvans:
- Undelete a version
- Hard delete a version
- Add logging for soft deleting

fixes #479

### DIFF
--- a/app/views/projects/versions/view.scala.html
+++ b/app/views/projects/versions/view.scala.html
@@ -12,6 +12,7 @@
 @import ore.permission.ViewLogs
 @import ore.Platforms
 @import ore.PlatformCategory
+@import ore.permission.HardRemoveVersion
 @(v: VersionData, sp: ScopedProjectData)(implicit messages: Messages, request: OreRequest[_], flash: Flash, config: OreConfig)
 
 @projectRoutes = @{controllers.project.routes.Projects}
@@ -135,6 +136,14 @@
                                     </button>
                                     <ul class="dropdown-menu" aria-labelledby="admin-version-actions">
                                         <li><a href="@appRoutes.showLog(None, None, None, v.v.id, None, None, None)">User Action Logs</a></li>
+                                        @if(request.data.globalPerm(ReviewProjects)) {
+                                            @if(v.v.visibility == VisibilityTypes.SoftDelete) {
+                                                <li><a href="#" data-toggle="modal" data-target="#modal-restore">Undo delete</a></li>
+                                            }
+                                            @if(request.data.globalPerm(HardRemoveVersion) && !v.isRecommended) {
+                                                <li><a href="#" data-toggle="modal" data-target="#modal-harddelete" style="color: darkred">Hard delete</a></li>
+                                            }
+                                        }
                                     </ul>
                                 </div>
                             }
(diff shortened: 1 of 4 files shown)


app/views/projects/versions/view.scala.html
```diff
@@ -135,6 +136,14 @@ <h1 class="pull-left">@v.v.versionString</h1>
                                     </button>
                                     <ul class="dropdown-menu" aria-labelledby="admin-version-actions">
                                         <li><a href="@appRoutes.showLog(None, None, None, v.v.id, None, None, None)">User Action Logs</a></li>
+                                        @if(request.data.globalPerm(ReviewProjects)) {
+                                            @if(v.v.visibility == VisibilityTypes.SoftDelete) {
+                                                <li><a href="#" data-toggle="modal" data-target="#modal-undelete">Undo delete</a></li>
+                                            }
+                                            @if(request.data.globalPerm(HardRemoveVersion) && !v.isRecommended) {
+                                                <li><a href="#" data-toggle="modal" data-target="#modal-harddelete" style="color: darkred">Hard delete</a></li>
```
phase: I believe we are actively against inline styling.

phase (commented):
Is `un-delete` a good word here? Maybe `restore`?

Katrix (approved):

app/views/projects/versions/view.scala.html
```diff
@@ -135,6 +136,14 @@ <h1 class="pull-left">@v.v.versionString</h1>
                                     </button>
                                     <ul class="dropdown-menu" aria-labelledby="admin-version-actions">
                                         <li><a href="@appRoutes.showLog(None, None, None, v.v.id, None, None, None)">User Action Logs</a></li>
+                                        @if(request.data.globalPerm(ReviewProjects)) {
+                                            @if(v.v.visibility == VisibilityTypes.SoftDelete) {
+                                                <li><a href="#" data-toggle="modal" data-target="#modal-undelete">Undo delete</a></li>
+                                            }
+                                            @if(request.data.globalPerm(HardRemoveVersion) && !v.isRecommended) {
```
mbax: Only folks with hard delete perm should be Ore admins, and they should be allowed to delete even if it's a recommended version.

mbax (changes requested):

app/views/projects/versions/view.scala.html
Ewoutvans: Its a bit more complicated to delete a recommended version the reason I did it like this is because a project must *always* have 1 public version (thats just the way Ore works) so to hard delete a version you should mark another build as recommended or soft delete it first (soft delete will set the second to last released version as recommended) and then hard delete. This forbids the bad practice of deleting the last public version (and breaking Ore) instead of removing the project.

mbax: But by the time you get to hard delete, you've already soft deleted, and thus the version should have been bumped from recommended?

phase: Yes, so this check is only a sanity check.

phase (approved):

PR #660, opened 2018-09-02
Use internal names instead of the forum role id

felixoi:


app/ore/permission/role/RoleTypes.scala
```diff
@@ -2,111 +2,54 @@ package ore.permission.role
...
-    * @param title  Title to display
-    * @param color  Color to display
-    */
-  class DonorType(override val i: Int,
-                  override val internalName: String,
-                  override val roleId: Int,
-                  override val title: String,
-                  override val color: Color)
-                  extends RoleType(i, internalName, roleId, classOf[GlobalRole], Default, title, color)
+sealed abstract class RoleType(val value: String ,val forumRoleId: Int, val roleClass: Class[_ <: Role], val trust: Trust,
```
Katrix: Can we make this a case class?

app/ore/permission/role/RoleTypes.scala
```diff
@@ -2,111 +2,54 @@ package ore.permission.role
...
+                               val title: String, val color: Color, val isAssignable: Boolean = true) extends StringEnumEntry with MappedType[RoleType] {
+  implicit val mapper: JdbcType[RoleType] = OrePostgresDriver.api.roleTypeTypeMapper
+}
 
-  implicit def convert(value: Value): RoleType = value.asInstanceOf[RoleType]
+sealed abstract class DonorType(override val value: String, override val forumRoleId: Int, override val title: String,
+                                override val color: Color, val rank: Int) extends RoleType(value, forumRoleId, classOf[GlobalRole], Default, title, color)
+
+case object RoleType extends StringEnum[RoleType] {
```
Katrix: Case object here is useless

build.sbt
```diff
@@ -25,6 +25,7 @@ libraryDependencies ++= Seq(
   "org.bouncycastle"      %   "bcpg-jdk15on"            %   "1.60",
   "javax.mail"            %   "mail"                    %   "1.4.7",
   "com.beachape"          %%  "enumeratum"              %   "1.5.13",
+  "com.beachape"          %%  "enumeratum-slick"        %   "1.5.15",
```
Katrix, line 28: Can you make sure the enumeratum and enumeratum-slick versions are in sync?

Katrix (changes requested):
Looks for the most part great (don't feel comfortable commenting on the evolutions)

build.sbt
felixoi: I think they are in sync. It uses the latest version of enumeratum and there is only one version of enumeratum-slick (which is 1.5.15).

Ewoutvans (changes requested):
Updating a project (adding member, changing role..)
Fails on the following line:

https://github.com/SpongePowered/Ore/blob/c0836e8afa8a034498555cef07570f3847bad66b/app/models/project/ProjectSettings.scala#L192

Ore receives the title not the internal_name and you are doing a search on Internal_name

`play.api.http.HttpErrorHandlerExceptions$$anon$1: Execution exception[[NoSuchElementException: Developer is not a member of ValueEnum (Ore_Admin, Ore_Mod, Sponge_Leader, Team_Leader, Community_Leader, Sponge_Staff, Sponge_Developer, Ore_Dev, Web_dev, Documenter, Support, Contributor, Advisor, Stone_Donor, Quartz_Donor, Iron_Donor, Gold_Donor, Diamond_Donor, Project_Owner, Project_Developer, Project_Editor, Project_Support, Organization, Organization_Owner, Organization_Admin, Organization_Developer, Organization_Editor, Organization_Support)]]`

app/db/impl/access/UserBase.scala
```diff
@@ -108,13 +107,13 @@ class UserBase(override val service: ModelService,
   def getStaff(ordering: String = ORDERING_ROLE, page: Int = 1)(implicit ec: ExecutionContext): Future[Seq[User]] = {
     // determine ordering
     val (sort, reverse) = if (ordering.startsWith("-")) (ordering.substring(1), false) else (ordering, true)
-    val staffRoles = List(RoleTypes.Admin, RoleTypes.Mod)
+    val staffRoles: List[RoleType] = List(RoleType.OreAdmin, RoleType.OreMod)
```
Katrix, line 110: Any reason for specifying the type here?

felixoi: Yes. I tried it without but it threw a compiler error since slick didn't knew which type to bind on with the `bind` method used below.

Ewoutvans (approved):

Katrix (approved):

PR #664, opened 2018-09-13
Improve performance in general, and especially on homepage

Katrix:
Improves speed in general, especially for the homepage.

Permissions now don't need n db trips to check n different permissions (not used in too many places currently). Also simplified permission stuff a bit. As a result, HeaderData should now be quicker to create.

Also added a few indexes and dropped an unneeded table.

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -63,7 +63,7 @@ final class Application @Inject()(data: DataHelper,
     Ok(views.linkout(remoteUrl))
   }
 
-  private def queryProjectRV = {
+  private val queryProjectRV = {
     val tableProject = TableQuery[ProjectTableMain]
     val tableVersion = TableQuery[VersionTable]
     val userTable = TableQuery[UserTable]
@@ -72,9 +72,7 @@ final class Application @Inject()(data: DataHelper,
       p <- tableProject
       v <- tableVersion if p.recommendedVersionId === v.id
       u <- userTable if p.userId === u.id
-    } yield {
-      (p, u, v)
-    }
+    } yield (p, u, v)
   }
 
   /**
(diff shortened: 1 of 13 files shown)


app/ore/permission/PermissionPredicate.scala
```diff
@@ -1,67 +1,27 @@
...
-        case true => Future.successful(!not)
-        case false =>
-          for {
-            result <- subject.scope.test(user, p)
-          } yield {
-            if (not) !result else result
-          }
-      }
+      if (user.globalRoles.contains(RoleType.OreDev) && (p == ViewHealth || p == ViewLogs || p == ViewActivity || p == ViewStats)) true
+      else p.trust <= trust
```
phase, line 23: Thank god this is finally gone :pray:

conf/evolutions/default/103.sql
```diff
@@ -0,0 +1,34 @@
+# --- !Ups
+DROP TABLE projects_deleted;
```
phase: We might want to keep a table of hard deleted projects just in case we need to reference them in something like the Moderation Log.

app/controllers/Application.scala
```diff
@@ -107,48 +105,41 @@ final class Application @Inject()(data: DataHelper,
...
-            projects zip tags map { case ((p, u, v), t) =>
-              (p, u, v, t)
-            }
-          }
-        }
-
-        queryProjects() map { data =>
-          val catList = if (categoryList.isEmpty || Categories.visible.toSet.equals(categoryList.toSet)) None else Some(categoryList)
-          Ok(views.home(data, catList, query.find(_.nonEmpty), p, ordering, pcat, pform))
+    val versionIdsOnPlatform = TableQuery[TagTable].filter(_.name.toLowerCase inSetBind platformNames).map(_.versionIds.unnest)
```
phase, line 109: :pray:

phase (approved):
I don't know what @mbax wants to do about logging hard deleted projects, but that can be dealt with later. This PR looks good :star:

conf/evolutions/default/103.sql
felixoi: I don't think so. Soft deleted projects will stay in the projects table and hard delete can only be done by admins and it should really remove all data then.

felixoi (approved):

PR #669, opened 2018-09-15
Add database tests and fix errors that cropped up

### LINKED ISSUE #583: Use doobie to test some of our SQL stuff
While I don't think we'll switch over to doobie for out main stuff, using it to test our SQL should be a viable thing to do.

When we do, doobie will tell us about how shit our DB and queries are, what types we should use instead, and what constraints we should use.

Katrix:
Adds the first tests to Ore.

Currently this mostly just covers everything that touches the database and object ids.

Resolves #583


Katrix:
Resolves #583

felixoi (approved):

PR #671, opened 2018-09-19
Use enumeratum

Katrix:
Replaces all usages of Scala's enumeration with enumeratum.

Note that in some cases when dealing with the DB, the enum type is specified to not get a too specific type.


felixoi (approved):
Just a fast review. Looks good so far but I didn't check every feature which could be impacted.

phase (approved):

PR #672, opened 2018-09-20
Misc cleanup and fixes

Katrix:
Adds both actual stuff that we need to fix (no vals in traits), and lot's of cleanup everywhere (especially controllers).

Assumes #671 is merged first.


felixoi (changes requested):
Downloads are broken for me I always get the warning "Could not verify your confirmation. It might have timed out, already been used, or something might have changed on your end". Everything else looks good.

felixoi (approved):

PR #674, opened 2018-10-01
Normalize db

Katrix:
Roles now exist in the db in their own table.
Tags no longer use an array for their versions.


app/db/impl/schema/tableSchema.scala
```diff
@@ -604,3 +600,39 @@ class LoggedActionViewTable(tag: RowTag) extends ModelTable[LoggedActionViewMode
     (ppId.?, ppSlug.?) <> ((LoggedProjectPage.apply _).tupled, LoggedProjectPage.unapply)
   def loggedSubjectProjection = (sId.?, sName.?) <> ((LoggedSubject.apply _).tupled, LoggedSubject.unapply)
 }
+
+class DbRoleTable(tag: Tag) extends ModelTable[DbRole](tag, "roles") {
```
Ewoutvans, line 604: Why the sudden use of the 'Db' prefix?

conf/evolutions/default/107.sql
```diff
@@ -0,0 +1,133 @@
...
+  category      ROLE_CATEGORY      NOT NULL,
+  trust         INTEGER            NOT NULL,
+  title         VARCHAR(255)       NOT NULL,
+  color         VARCHAR(255)       NOT NULL,
+  is_assignable BOOLEAN            NOT NULL,
+  rank          INTEGER
+);
+
+INSERT INTO roles (id, name, category, trust, title, color, is_assignable, rank)
+VALUES (61, 'Ore_Admin', 'global', 5, 'Ore Admin', '#DC0000', TRUE, NULL),
```
Ewoutvans: Can we ditch the id's here?
I don't think it's wise to keep prolonging life to the old forumId system and if we migrate them it will only make it harder to remove it in the future.

@felixoi what do you think we should do.

conf/evolutions/default/107.sql
```diff
@@ -0,0 +1,133 @@
...
+
+INSERT INTO roles (id, name, category, trust, title, color, is_assignable)
+VALUES (64, 'Organization', 'organization', 5, 'Organization', '#B400FF', FALSE),
+       (-5, 'Organization_Owner', 'organization', 5, 'Owner', '#B400FF', FALSE),
+       (-9, 'Organization_Admin', 'organization', 4, 'Admin', '#B400FF', TRUE),
+       (-6, 'Organization_Developer', 'organization', 3, 'Developer', 'transparent', TRUE),
+       (-7, 'Organization_Editor', 'organization', 1, 'Editor', 'transparent', TRUE),
+       (-8, 'Organization_Support', 'organization', 0, 'Support', 'transparent', TRUE);
+
+CREATE INDEX role_name_idx ON roles (name);
```
Ewoutvans: I think this should be a unique index (I don't know if SpongeAuth forced uniqueness should probably be checked)

Ewoutvans (changes requested):

app/db/impl/schema/tableSchema.scala
Katrix: Just to try to stay consistent with what appears to be the general rule for naming the tables (<ModelName>Table). Don't really think the table will see much use from Slick directly anyway.

felixoi (approved):

Ewoutvans (approved):

felixoi (changes requested):
Tags are broken when using this PR.
It seems like you break it in the db since tags are still broken when switching back to master with the same database.

phase (commented):
:clap: The version tags have been needing this for a while.

app/db/impl/schema/tableSchema.scala
felixoi: Maybe rename it to `GlobalRoleTable`?

Katrix: 1. that doesn't really describe what they are as they can be used in projects and orgs too. 2. I can't wait for one of use to ask the other what was the difference between `GlobalRolesTable` and `UserGlobalRolesTable`

felixoi (approved):

PR #678, opened 2018-10-04
Make having a scope a typeclass property

Katrix:
Removes `ScopeSubject` and replaces it with a typeclass `HasScope`. In essence, 
`def foo(subject: ScopeSubject): Bar` 
becomes 
`def foo[A: HasScope](subject: A): Foo`


phase (approved):
👍 just rebase on master and we're good to go

felixoi (approved):

PR #679, opened 2018-10-05
Only run task on visible projects and add some more sentry stuff

Ewoutvans:


Katrix (changes requested):
Fix the imports. You're removing newlines which should be there

app/discourse/RecoveryTask.scala
```diff
@@ -36,12 +37,15 @@ class RecoveryTask(scheduler: Scheduler, retryRate: FiniteDuration, api: OreDisc
   override def run(): Unit = {
     Logger.debug("Running Discourse recovery task...")
 
-    this.projects.filter(_.topicId === -1).foreach { toCreate =>
+    val topicFilter: ModelFilter[Project] = ModelFilter[Project](_.topicId === -1)
```
felixoi: This will never be -1 anymore, it's null if no topic is created.

felixoi (changes requested):

felixoi (approved):

PR #687, opened 2018-10-12
Fix upload when using organization as owner

### LINKED ISSUE #681: Attempt to create a new plugin under an org results in inconsistent state
I attempted to create a new plugin under the Nucleus namespace, Heisenberg. During the project creation, I set that the owner of the plugin should be the Nucleus org, which I own. I expected that the plugin would be created, and that I would be able to alter the documentation and upload new version as required.

However, this has not happened, instead, I've ended up with a plugin in a very weird state:

* The plugin can be seen under the org page at https://ore.spongepowered.org/Nucleus
* The plugin exists in the `dualspiral` namespace (see https://ore.spongepowered.org/dualspiral/Nucleus-Heisenberg)
* I have no editing or upload rights, as I am not listed as an owner of the plugin

I would expect that the plugin would be under the `Nucleus` namespace and that as the owner of the Nucleus org, I would inherit the permissions from the org with regards to this project unless otherwise specified.

This is using the latest version of Ore as of the time of writing, 7th Oct 2018 (which might be 1.6.14, but I'm not sure if there has been a deploy of this version)

felixoi:
This is not the best fix but it works and because no one has time to fix this properly we should go with this for now. Since upload is a key feature of Ore we should get this up afap.

Fixes #681

### DIFF
--- a/app/models/project/ProjectSettings.scala
+++ b/app/models/project/ProjectSettings.scala
@@ -68,13 +68,22 @@ case class ProjectSettings(
     Logger.debug("Saving project settings")
     Logger.debug(formData.toString)
 
-    val updateProject = service.updateIfDefined(
-      project.copy(
-        category = Category.values.find(_.title == formData.categoryName).get,
-        description = noneIfEmpty(formData.description),
-        ownerId = formData.ownerId.getOrElse(project.ownerId)
+    val queryOwnerName = for {
+      u <- TableQuery[UserTable] if formData.ownerId.getOrElse(project.ownerId).bind === u.id
+    } yield {
+      u.name
+    }
+
+    val updateProject = service.doAction(queryOwnerName.result).flatMap { ownerName =>
+      service.updateIfDefined(
+        project.copy(
+          category = Category.values.find(_.title == formData.categoryName).get,
+          description = noneIfEmpty(formData.description),
+          ownerId = formData.ownerId.getOrElse(project.ownerId),
+          ownerName = ownerName.head
+        )
       )
-    )
+    }
 
     val updateSettings = service.updateIfDefined(
       copy(
--- a/app/models/user/User.scala
+++ b/app/models/user/User.scala
@@ -10,16 +10,8 @@ import play.api.mvc.Request
 import db.access.{ModelAccess, ModelAssociationAccess}
 import db.impl.OrePostgresDriver.api._
 import db.impl.access.{OrganizationBase, UserBase}
-import db.impl.schema.{
-  OrganizationMembersTable,
-  OrganizationRoleTable,
-  OrganizationTable,
-  ProjectMembersTable,
-  ProjectStarsTable,
-  ProjectWatchersTable,
-  UserTable
-}
-import db.{Model, ModelService, Named, ObjectId, ObjectReference, ObjectTimestamp}
+import db.impl.schema._
+import db._
 import models.project.{Flag, Project, Visibility}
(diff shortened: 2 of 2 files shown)


app/controllers/project/Projects.scala
```diff
@@ -1,16 +1,11 @@
...
-
-import scala.collection.JavaConverters._
-import scala.concurrent.{ExecutionContext, Future}
-
-import play.api.cache.AsyncCacheApi
-import play.api.i18n.MessagesApi
-import play.api.libs.Files.TemporaryFile
-import play.api.mvc.{Action, AnyContent, MultipartFormData, Result}
 
+import _root_.util.StringUtils._
```
phase: @Katrix How do you want to deal with import order?

app/models/project/ProjectSettings.scala
```diff
@@ -70,13 +70,22 @@ case class ProjectSettings(
...
+      u.name
+    }
+
+    val updateProject = service.doAction(queryOwnerName.result).flatMap { ownerName =>
+      updateIfDefined(
+        project.copy(
+          category = Category.values.find(_.title == formData.categoryName).get,
+          description = noneIfEmpty(formData.description),
+          ownerId = formData.ownerId.getOrElse(project.ownerId),
+          ownerName = ownerName.head
```
phase, line 83: Was this the main bug? We really need to gut names out soon.

app/models/user/User.scala
```diff
@@ -141,22 +130,25 @@ case class User(
...
-                  m  <- memberTable if m.organizationId === o.id
-                  r  <- roleTable if m.userId === r.userId && r.organizationId === o.id
+                  p <- projectsTable if pScope.projectId.bind === p.id
+                  o <- orgaTable if p.userId === o.id
+                  m <- memberTable if m.organizationId === o.id && m.userId === this.userId.bind
+                  r <- roleTable if this.userId.bind === r.userId && r.organizationId === o.id
                 } yield r.roleType
 
-                service.doAction(query.to[Set].result).map(biggestRoleTpe)
+                service.doAction(query.to[Set].result).map { roleType =>
```
phase: This variable should be called `roleTypes` since it's a list.

app/models/user/User.scala
```diff
@@ -141,22 +130,25 @@ case class User(
...
+                  if (roleType.contains(RoleType.OrganizationAdmin) || roleType.contains(RoleType.OrganizationOwner)) {
```
phase: Why are we not always calling `biggestRoleType` here? Shouldn't that method be the one that checks for Org roles?

phase (commented):
Ref #681

app/models/project/ProjectSettings.scala
felixoi: This was the bug which created the project under the wrong namespace, yes.

app/controllers/project/Projects.scala
Katrix: It's all in the contributing document.  https://github.com/SpongePowered/Ore/blob/master/CONTRIBUTING.md#import-formatting

app/models/user/User.scala
felixoi: No. We only want to inherit permissions from orgs if it's an orga owner or an orga admin. `biggestRoleTpe` is a method on the `User` class which gets the role with the highest trust, it's more general and not only for use in this case.

phase (approved):

PR #688, opened 2018-10-14
Only update version if it's defined

Katrix:
Fixes api deploy with a changelog.


Ewoutvans (approved):
API uploads work with changelog

felixoi (approved):

PR #690, opened 2018-10-15
Remove model schema

Katrix:
Or, "Create a typeclass for getting the base Query for a model"

`ModelSchema` and its friends have been a big thorn in my side when trying to improve the db code. It turns out, the two big things it's used for is storing the base query, and how to get to children. This PR introduces a typeclass for getting the base query. Getting the children is now handled manually. In the few cases where something else in it was used, I dragged the code out the use site.


Katrix:
Superseded by #696

PR #695, opened 2018-10-26
Make it possible for a project to have no recommended version

Katrix:
Does what it says.

This was a surprisingly simple change.


build.sbt
```diff
@@ -13,7 +13,7 @@ scalacOptions ++= Seq(
   "-feature",
   "-unchecked",
   "-Xcheckinit",
-  "-Xfatal-warnings",
+  //"-Xfatal-warnings",
```
felixoi: Intended?

Katrix: Oops, will fix

PR #696, opened 2018-10-26
Misc cleanup for db code

Katrix:
To be merged after #690.

Includes misc improvements to the db code.
* copyWith moved to ModelQuery typeclass. No more casts around that because of that.
* ModelService is now the only way to execute DBIO actions
* ModelService#doAction is now called runDBIO, to it clearer once doobie can also only run it's programs from here.
* Removed some unused classes and members
* ModelBases are now gotten statically from ModelService
* ModelFilter is now what it was likely what it was supposed to be, a helper for creating `M#T => Rep[Boolean]` values. It now acts as a proper ops value class. It is no longer taken as a parameter anywhere because if this.
* Removed usage of ModelFilter where it is no longer needed.
* Made a few calls that used ModelAccess#all instead do more stuff in the db
* More stuff I've probably forgotten


Katrix:
Superseded by typed-ref branch.

PR #702, opened 2018-11-06
Clean up DB code

Katrix:
I don't even know what to call this anymore. In general, this PR just improves the code around the DB. First, let's go over the features that were carried over from the two PRs it superseded.

* Delete `ModelSchema` and co
* Create a typeclass for getting the query for a type. This typeclass also handles the `copyWith` method now.
* `ModelService` is now the old way to interact with the DB.
* `ModelService#doAction` is now called `runDBIO` to better reflect it's new companion, `runConIO` which runs a doobie program.
* `ModelBases` are now gotten statically from `ModelService`.
* `ModelFilter` is now what it was likely what it was supposed to be, a helper for creating `M#T => Rep[Boolean]` values. It now acts as a proper ops value class. It is no longer taken as a parameter anywhere because if this.
* Removed usage of `ModelFilter` where it is no longer needed.
* Removed null usage in `ModelService`.
* Made a few calls that used `ModelAccess#all` instead do more stuff in the db.
* More stuff I've probably forgotten

And then on to the new stuff
* Created a better `ModelAssociationAccess` which serves most needs
* `ObjectReference` -> `DbRef[A]`
* AssociativeTable now remembers it's type information
* Split up the table schema file into individual files. This should help with both incremental compilation and hydra performance
* More stuff I've probably forgotten


app/db/impl/OrePostgresDriver.scala
```diff
@@ -57,6 +57,12 @@ trait OrePostgresDriver
     implicit val langTypeMapper: BaseColumnType[Lang] =
       MappedJdbcType.base[Lang, String](_.toLocale.toLanguageTag, Lang.apply)
 
+    /*
+    implicit def dbRefBaseType[A]: BaseColumnType[DbRef[A]] = longColumnType.asInstanceOf[BaseColumnType[DbRef[A]]]
+    implicit def dbRefArrayBaseType[A]: BaseColumnType[List[DbRef[A]]] =
+      simpleLongListTypeMapper.asInstanceOf[BaseColumnType[List[DbRef[A]]]]
+     */
```
felixoi, line 66: Why you are commenting this out?

app/db/package.scala
```diff
@@ -1,5 +1,5 @@
 package object db {
 
-  type ObjectReference = Long
-
+  //type DbRef[A] = Long @@ RefLink[A]
```
felixoi, line 3: Is this for later use?

app/db/impl/OrePostgresDriver.scala
Katrix: For later use

app/db/package.scala
Katrix: Yes

PR #708, opened 2018-11-15
Delete unsafe download

Katrix:
This is apparently unused. We just insert into the table and delete the value after 30 days.


Ewoutvans:
What about the stats page? This metric is important so we know the ratio of downloads that are safe / unsafe.

Katrix:
Ah shoot, didn't think of that one. Forgot we still had raw SQL not in the queries objects. Guess I'll close this then. Still though, there is a lot more information in the unsafe downloads column that I can't really see us having a use for. Should we at least remove that stuff?

PR #709, opened 2018-11-15
Delete version assets

Katrix:
I have no idea what this was supposed to be used for, but it's never been used, and all versions currently have it set to null.


phase:
[This commit](https://github.com/SpongePowered/Ore/commit/8078297d99a588019f4a457a0ec244abb39bfaec#diff-b551087a4caf7978dd07b350868de9afR50) added that field. No clue what the intent was.

Issue #568, opened 2018-07-15
Setup project redirects back to Upload project if project owner is set to be organization


HaedHutner:
Browser & Version: `Google Chrome | 67.0.3396.99 (Official Build) (64-bit) (cohort: Stable)`

Operating System: `Windows 7 64-bit`

Error message (if applicable): `N/A`

Steps to reproduce:

* Start creation of new project on ore
* Select and upload signed plugin jar
* On the `Setup project` stage, change `Owner` from the default ( which is the user ), to an organization
* Fill out rest of form appropriately
* Click `Continue` at the bottom
* It brings you back to the `Upload project` stage

phase:
The project was also a deleted project that the user was trying to recreate.

Issue #558, opened 2018-07-05
Creating a new project resets to beginning after "invite members" page


parlough:
Browser & Version: Chrome Canary - Version 69.0.3481.0 (Official Build) canary (64-bit)

Operating System: macOS High Sierra - Version 10.13.6 Beta and Windows 10 Build 17692

Error message (if applicable): While I am uploading a new project, after the invite members page, I get sent to the beginning of project creation again.

From here:
<img width="1493" alt="screen shot 2018-07-04 at 8 09 21 pm" src="https://user-images.githubusercontent.com/18372958/42297651-93f24e6e-7fc6-11e8-9d6f-5f45cb9f9fdf.png">
To here:
<img width="1473" alt="screen shot 2018-07-04 at 8 09 31 pm" src="https://user-images.githubusercontent.com/18372958/42297653-9585b72a-7fc6-11e8-9ea9-d77fc253d28e.png">

Steps to reproduce:

* Go to create a project
* Upload jar and sig
* Fill up necessary information(I chose IchorPowered as owner)
* Go to add member page(either add somebody or don't and click next)

The project I was trying to upload is here - https://github.com/ichorpowered/eggers/tree/v0.1.0

Ewoutvans:
Can confirm that this is an issue; also takes ages to load for some reason

As you can see the members also shows my private account as owner, this should be the Organization.
Demo:
![kapture 2018-07-17 at 19 20 37](https://user-images.githubusercontent.com/3961526/42834155-923d1c04-89f6-11e8-9e9f-398baa1860e1.gif)

Ewoutvans:
After long time debugging I managed to figure out that the moment you press the button, https://github.com/SpongePowered/Ore/blob/c661b29b8e6e48709d2992333f49fea8da32849c/app/controllers/project/Projects.scala#L193
gets called but. For some reason the factory still thinks is under the key `username/slug` (username being the one who is logged in) and not `organization/slug`

PR #722, opened 2018-12-20
Add scalafix and fix most errors

Katrix:
Adds Scalafix, and fixes most warnings that it emits when running it. Won't update documentation just yet as there are still warnings left that I want to fix. Those will have to wait for another time though. 

I want to look around for some other stuff, and wait for some stuff before fixing those. Most notably I want the upload process changes in the bootstrap branch as I've already made quite a few changes in this area, and I don't want it to be too hard to merge in these changes.


.travis.yml
```diff
@@ -27,6 +27,7 @@ cache:
   directories:
     - $HOME/.ivy2/cache
     - $HOME/.sbt
+    - $HOME/.coursier/cache
```
felixoi, line 30: We are not using coursier in this repo. Why we should cache it then?

Katrix: We totally are. I just didn't know which directory to cache. Every time scalafmt runs it downloads stuff using coursier. Scalafix itself also uses coursier to download stuff.

Katrix:
Fixed some bugs in the PR, in addition to #558 and maybe #568.

app/security/pgp/PGPPublicKeyInfo.scala
```diff
@@ -37,93 +40,75 @@ object PGPPublicKeyInfo {
...
 
-          Logger.debug("Signatures:")
-          val sigIter = key.getSignatures
-          while (sigIter.hasNext) {
-            val sig: PGPSignature = sigIter.next().asInstanceOf[PGPSignature]
-            Logger.debug("\tCreated at: " + sig.getCreationTime)
-            Logger.debug("\tCertification: " + sig.isCertification)
-          }
+                Logger.debug("Users:")
+                val firstUser = key.getUserIDs.next()
```
felixoi: Throws an NullPointerException when adding a new key

felixoi (approved):

PR #724, opened 2019-01-11
Add proper MDC Resolves #215

### LINKED ISSUE #215: Add some MDC tags
Sentry is more useful when there's some context provided, like the username/user ID experiencing the problem, as well as the current version of Ore.

In `raven-logback` this is done by setting values in the [Mapped Diagnostic Context](https://logback.qos.ch/manual/mdc.html) - you don't get the full user context support 😢 but it's better than nothing.

You can also add some breadcrumbs, but that might be a liiiittle tedious.

Katrix:
Adds proper MDC passed as implicit values using scala-logging. In comparison to the previous MDC we had, this works across threads, meaning it won't maybe disappear the moment we do a map or a flatMap on an IO. As a bonus, scala-logging is also slightly faster than the logging provided by Play.

Closes #215

### DIFF
--- a/app/Bootstrap.scala
+++ b/app/Bootstrap.scala
@@ -8,6 +8,7 @@ import discourse.OreDiscourseApi
 import ore.OreConfig
 import ore.project.ProjectTask
 
+import com.typesafe.scalalogging
 import org.bouncycastle.jce.provider.BouncyCastleProvider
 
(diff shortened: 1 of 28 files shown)


felixoi (approved):
Looks good, please rename the stuff mentioned.

PR #731, opened 2019-01-15
Some changes in the base layout

felixoi:
Stuff fixed here:
- Javascript is always added at the end of `body`
- The base container was moved to the base layout and always has the top margin when the nav is shown
- Alerts were moved to base layout, this allows to redirect with error in every template

Includes a fix for #732.

### DIFF
--- a/app/views/utils/alert.scala.html
+++ b/app/views/utils/alert.scala.html
@@ -3,7 +3,9 @@
 @bootstrapNames = @{
     Map(
         "error" -> "alert-danger",
-        "success" -> "alert-success"
+        "success" -> "alert-success",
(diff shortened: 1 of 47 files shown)


app/views/utils/alert.scala.html
```diff
@@ -3,7 +3,9 @@
 @bootstrapNames = @{
     Map(
         "error" -> "alert-danger",
-        "success" -> "alert-success"
+        "success" -> "alert-success",
```
Katrix, line 6: If we now support more alert features, can you provide shorthand methods for them similar to `withError` and the others?

Katrix (approved):

PR #733, opened 2019-01-17
Changes to the logback configuration

felixoi:
- Disables the security filter since it only produces warnings which lead to nothing in production. See https://sentry.io/sponge/ore/?query=logback-Marker%3A%22SECURITY%22
- Adds graceful shutdown for logback

### DIFF
--- a/conf/logback.xml
+++ b/conf/logback.xml
@@ -15,6 +15,11 @@
         <appender-ref ref="Console" />
     </appender>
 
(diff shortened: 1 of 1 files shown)


Katrix:
Does the shutdown stuff work with reloading code in dev?

Katrix (approved):
Seems to work fine, and is also recommended by the play docs regarding the shutdown hook.

Issue #556, opened 2018-06-27
Remove ORMish DB code and replace it with something better


Katrix:
Creating this issue to gather discussion on the current state of the DB code in Ore, and possible improvements to it.

As I see it, the problems with the current code for handling the DB stuff are
1. Uses projection types, which won't be around forever, and they also work poorly with IntelliJ
2. Mutable case classes
3. "Fake" Option id values
4. Very tight coupling between the models and the DB
5. Class objects everywhere
6. Many parts are dynamic and can fail at runtime
7. Encourages too much code reuse instead of writing new/better queries

I think the best approach would be to move the majority of the DB accessors and similar to typeclasses which use dependent types instead. These typeclasses should be the only way to get or create the models. That would solve 1, 2, 4, 5 and 6, and partially solve 3. For the rest of 3, let's just create an `ObjectId` type which has a single public value outside the db package, which is `value`. It will act like Option and on the inside, but that should not be exposed to the outside as such a value should never escape to the outside.

Sadly I don't currently have any pseudocode to show, but if anyone is interested I can probably write some up.

Note that solving this issue would take a lot of code changes and should be done incrementally. For example, ObjectId is probably a nice place to start and should be its own PR.

Katrix:
So, I've had some time to think of how to solve this issue, so I'll ramble down here a bit showing some code, and my current plans. If anyone has any better ideas or disagree with my thoughts, feel free to say so.

First, let's talk about the amount of code reuse that's currently happening. My proposed solution is that we need some type `Q[A]` which lets us do operations like filtering and sorting as chained methods. The type should not do anything unless it's interpreted. Luckily for us, Slick already has something like that called `Query`. The majority of methods on ModelAccess and friends should be removed or changed to return `Query` as you might want to do something else on the query they use afterward.

While we're at it, we should probably move from `Future` to `DBIO` in some cases. Do note however that `DBIO` always executes sequentially. 

Next, let's talk about the DB models, how to insert update and all that.

First of, the model type.
```scala
case class Model[A](
  id: ObjectId[A],
  createdAt: ObjectTimestamp,
  value: A
)
```

Here the model's value, and it's DB properties are correctly separated. While we probably won't get completely there, the value should know as little as possible about the DB.

`ObjectReference` and `ObjectId` also take a type parameter describing what type they reference. These will probably be aliased to avoid boilerplate.

`Channel` would, for example, look like this
```scala
case class Channel(
[...]

PR #734, opened 2019-01-18
Remove/Redo ORMish DB code in Ore

Katrix:
Some misc db improvements.

* Introduced methods on ModelService and ModelAccess that returns Query and Rep and renamed existing methods to have a `now` suffix
* ObjId and ObjTimestamp are now impossible to access outside the db package. As such I've defined an implicit conversion from ObjId to DbRef. Bit iffy on this one. In some cases it works, in other cases it does not. Cases it does not work are mainly comparisions and places where an implicit conversion is already used.
* Do some more work in the db, and merge some db calls.
* Removed some unused stuff on ModelService. The db default timeout config value. Removed the start method on ModelService as it no longer carries state.


Katrix:
Adding a note here now because I think this PR will work (unlike my other attempts). This PR will resolve #556

build.sbt
```diff
@@ -13,7 +13,7 @@ scalacOptions ++= Seq(
   "-feature",
   "-unchecked",
   "-Xcheckinit",
-  "-Xfatal-warnings",
+  //"-Xfatal-warnings",
```
felixoi: Should this be enabled again?

felixoi (approved):

PR #735, opened 2019-01-19
Application to raw sql

Katrix:
Rewrites some view routes to raw sql and optimizes that in the process. Also introduces views to do trust calculation in the db. Lastly, switches over to full text search for plugin search. Should be both faster and better. This PR requires a postgres version later than 11.

Still slightly work in progress.


progwml6:
Before this is merged, I want a summary on how you plan on avoiding SQL injection when writing manual queries

Katrix:
Library handles it. `sql"SELECT t.value FROM table t WHERE t.str = $variable"` basically becomes (psudocode) `Fragment("SELECT t.value FROM table t WHERE t.str = ? ",  Seq(varable))`. There is a escape hatch function called `Fragment.const(str)` that works basically the same as `@Html(str)` in the views, but it's not used anywhere here.

progwml6:
this should be sufficient

app/db/query/AppQueries.scala
```diff
@@ -0,0 +1,251 @@
...
+    val frags = sql"SELECT * FROM v_logged_actions la " ++ Fragments.whereAndOpt(
+      userFilter.map(id => fr"la.user_id = $id"),
+      projectFilter.map(id => fr"la.filter_project = $id"),
+      versionFilter.map(id => fr"la.filter_version = $id"),
+      pageFilter.map(id => fr"la.filter_page = $id"),
+      actionFilter.map(i => fr"la.filter_action = $i"),
+      subjectFilter.map(id => fr"la.filter_subject = $id")
+    ) ++ fr"ORDER BY la.id DESC OFFSET $offset LIMIT $pageSize"
+
+    println(frags.query[LoggedActionViewModel[Any]].sql)
```
felixoi: Debug code?

Ewoutvans (approved):
Trust system looks good

felixoi (approved):

PR #740, opened 2019-01-30
Extend issue templates

felixoi:
Replaces the current issue template with one for bugs and one for feature requests.
Also enables automatic labeling if used. They are nearly the default ones but IMO they fit very good.


phase (approved):
:clap: I like these a lot

Katrix (approved):

PR #746, opened 2019-02-15
Update to play 2.7

Katrix:
Does as it says. Homepage seems to have loaded fine. Currently dummied out forums stuff. This also replaces our own CSP stuff with Play's CSP stuff.


app/util/StringUtils.scala
```diff
@@ -79,4 +80,22 @@ object StringUtils {
     */
   def prettifyDateAndTime(date: Date)(implicit messages: Messages): String =
     DateFormat.getDateTimeInstance(DateFormat.DEFAULT, DateFormat.DEFAULT, messages.lang.locale).format(date)
+
+  //https://stackoverflow.com/a/9855338
+  private val hexArray = "0123456789ABCDEF".toCharArray
+  def bytesToHex(bytes: Array[Byte]): String = {
```
kashike, line 86: Can't you do https://stackoverflow.com/a/2756627 here?

Katrix: Holy shit, a Kashike :P
Anyway, from my own guesses, and from the comments seen here, because it's much slower https://stackoverflow.com/a/13006907/7207457

felixoi (approved):

PR #747, opened 2019-02-16
only show pgp modal with private info to authenticated users

Ewoutvans:


Katrix:
The raw key is still visible to all right?

Ewoutvans:
@Katrix yes

Katrix (approved):

PR #749, opened 2019-02-16
Send version description in API

### LINKED ISSUE #745: Add changelog to API
**Describe the solution you'd like**
Currently on the API requests for the [project](https://docs.spongepowered.org/stable/en/ore/routes/project.html) and the [versions](https://docs.spongepowered.org/stable/en/ore/routes/list-versions.html) the changelog is currently not available. It would be nice to be able to get it through the API.

**Describe alternatives you've considered (optional)**
I'd personally prefer if the changelog info would included in the "version block". But a separate API request for a specific version would be accetable too.

**Additional context**
I'm working on an update checker plugin and would like to be able to display the changelog under certain conditions.

Katrix:
Sends a versions description in the API.
Closes #745

### DIFF
--- a/app/ore/rest/OreRestfulApi.scala
+++ b/app/ore/rest/OreRestfulApi.scala
@@ -176,7 +176,8 @@ trait OreRestfulApi extends OreWrites {
       "reviewState"   -> v.reviewState.toString,
       "href"          -> ('/' + v.url(p)),
       "tags"          -> tags.map(toJson(_)),
-      "downloads"     -> v.downloadCount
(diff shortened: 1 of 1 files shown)


BrainStone:
Isn't `v.description` an optional value?

I'm not too familiar with scala, so apologies if that's something stupid to point out.

Katrix:
It'll become null in the Json.

BrainStone:
Ah ok. Awesome. If I knew that, I'd made the PR myself xD

felixoi (approved):

PR #755, opened 2019-02-20
Separate project and version creation

Katrix:
Project and version creation are now separate things. Now you create a new project and upload the first version like any other.


Katrix:
This should be somewhat ready now

PR #775, opened 2019-03-18
Fix and simplify starring #Fixes 762

Katrix:
Fix and simplify starring #Fixes 762. Starring is now a toggle instead of set true or false.

### DIFF
--- a/app/controllers/project/Projects.scala
+++ b/app/controllers/project/Projects.scala
@@ -420,10 +420,11 @@ class Projects @Inject()(stats: StatTracker, forms: OreForms, factory: ProjectFa
     * @param starred True if should set to starred
     * @return Result code
     */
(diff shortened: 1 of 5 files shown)


app/controllers/project/Projects.scala
```diff
@@ -420,10 +420,11 @@ class Projects @Inject()(stats: StatTracker, forms: OreForms, factory: ProjectFa
     * @param starred True if should set to starred
     * @return Result code
     */
-  def setStarred(author: String, slug: String, starred: Boolean): Action[AnyContent] =
+  def toggleStarred(author: String, slug: String): Action[AnyContent] =
     AuthedProjectAction(author, slug).asyncF { implicit request =>
+      println("Foo")
```
Ewoutvans: Foo!

Ewoutvans (approved):

PR #784, opened 2019-04-03
API V2

### LINKED ISSUE #439: API rename, and add scoping controls, and owner/organization level keys
Rename Deployment Key field to Project API key, add User and Organization level keys as well(keep project level keys too).

For each type of key, there should be an ability to lock down the function of the keys for security reasons, examples of functionality that needs to be added:
* File upload
* Changelog upload
* Wiki upload/delete/change

### LINKED ISSUE #736: DateTimes returned from the API should be in ISO 8601 timestamp format
As discussed with felixoi and Katrix, it seems that currently date-times are not formatted by Ore itself, but just uses the PostgreSQL format. Ore should standardize date-times on the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format as it is the most widely used standard across the web.

Katrix:
Let's actually create a PR for this. Still not done.

After this PR, we need to enable the extension pgcrypto.

Closes #439
Fixes #736


public/javascripts/versionList.js
```diff
@@ -123,112 +127,142 @@ function loadVersions(increment, scrollTop) {
...
+        page += increment;
 
-            // Sets the new page number
-            page += increment;
-            var totalPages = Math.ceil(TOTAL_VERSIONS / VERSIONS_PER_PAGE);
+        totalVersionsPromise.then(function(totalVersionsResponse) {
+            totalVersions = totalVersionsResponse.count;
+            var totalPages = Math.ceil(totalVersions / VERSIONS_PER_PAGE);
 
+            console.log(totalVersionsResponse);
```
felixoi: Remove debug from here ongoing

felixoi (changes requested):

app/views/users/apiKeys.scala.html
```diff
@@ -0,0 +1,76 @@
...
+}
+
+@users.view(u, o, additionalScripts = scripts) {
+    <div class="row">
+        <div class="col-md-6">
+            <h2>Create new key</h2>
+            <div id="api-create-key-form">
+                <div class="row">
+                    <div class="col-md-6">
+                    @for(perm <- NamedPermission.values.take(NamedPermission.values.length / 2)) {
```
felixoi: Please modify this to only show the permissions the user has access to.

PR #786, opened 2019-04-06
Purge PGP from the Ore codebase

### LINKED ISSUE #159: Add option to pull PGP public key from keybase.io
This just requires pulling the key from the user's profile (e.g. https://keybase.io/windy/key.asc).

Parent issue: #151

### LINKED ISSUE #267: 3 days to submit after public key addition.
When you change your public key, you must wait 3 days to submit a plugin. Not a big deal, but was wondering if this was intended to be applied to those who have yet to submit a plugin?

ie: New to ore, adds public key, changes public key(or not- does this happen even if you just add and not change?), goes to submit- needs to wait 3 days.

Now, if this was the intended behaviour- understandable. Any plans to reduce this to 1 day for anyone who is submitting their first plugin?

Thanks in advance!

### LINKED ISSUE #318: Whose signature should be used on upload?
It's currently possible (as a staff role) to upload to other peoples projects, with #302 being merged, this is now (somewhat) transparent to the end user. (I'd argue uploads with signatures from those other then the **current** project team, should be highlighted as red to indicate error, red for ex-members and staff uploading on someones behalf to make sure they standout)

However, during upload, it's impossible to upload on another users behalf (currently, on the web frontend)

If the binaries have been signed off by an author, I see no reason why staff, or an organization account (used for a CI server or something strange?) shouldn't be able to upload using a provided signature, and binary (not key) for another author.

Maybe there is some security aspect I'm not considering, (I suspect it was originally done as a secondary guard against key theft).

 But is this functionality wanted?

I ran into it when uploading a plugin on behalf of Phit, when investigating the missing jar issue, to work out if it was possibly a sketchy internet connection, but found I had to sign with my own key, rather then using phits signature he sent me.

I would have thought, once signed it would have been all good?

### LINKED ISSUE #497: Multiple PGP keys per user
Would it be feasible to add multiple PGP keys to a user? I'd like to use oreDeploy to push CI builds from my CI server to my project, but I do not want to use my own PGP key to sign those builds because I don't want my private key on that server, just in case I do something really stupid on there and accidentally expose it. That way I can still create releases and just revoke my CI key if I need to.

This should, of course, be subject to the three day rule, and the user should be notified though e-mail that a key was added.

### LINKED ISSUE #590: Selectively improved response to incorrect signature upload
Proposing responses in these special cases:

* Public key uploaded instead: Re-explain they need to sign it, direct to documentation (which is great).
* Private key uploaded instead: Mark key as compromised, disallow further use by any user?

### LINKED ISSUE #758: HTML printed literally in error message when trying to create a project when a user doesn't have a public key
Since #512, HTML is always escaped in alerts. 

This message contains an HTML link:
https://github.com/SpongePowered/Ore/blob/436e7afece95b322d7c332f82159a0e3d06b810e/conf/messages#L101

image

Possible resolutions:

- Escape any unsafe input before passing to the template, then printing raw HTML in the template
- Have some sort of string replacement like `[LINK]...[/LINK]` that gets converted to `<a>` tags

Katrix:
To be merged after #784 

Not sure if I got everything, of it something is missing. Also fixes the deploy API to work now.

Seeing as we're removing them, I guess this
Closes #758 
Closes #590 
Closed #497 
Closed #318 
Closes #267 
Closes #159


simon816:
Has there been any discussion on this. I am not in favour of removing signing/integrity checking all together. Is there a proposed replacement?

Grinch:
This was discussed among the Ore staff members (dev and community) and decided upon. There isn't truly many benefits from the system as it stands and just serves to add an obstacle to folks wanting to simply upload something.

NeumimTo:
I suppose it would be enough to make this feature optional rather than removing it completly .  On the project page let the owner deceide whenever all uploads has to be signed

Katrix:
We did agree when we talked about this to remove it, not just make it optional, as it is in it's current form fundamentally flawed. Yes, we did consider making it optional but decided against it.

This is my own opinion, but if we do bring back PGP stuff, I don't want it to be like this. If we do bring it back, I want to make sure it actually provides more security, and people with experience around this can tell me "yes, this will provide more security, and isn't just for show". As it is right now it's just an inconvenience without too much gain.

felixoi (approved):

PR #787, opened 2019-04-10
Oops, we are in 2019 already

felixoi:


app/views/bootstrap/footer.scala.html
```diff
@@ -41,7 +41,7 @@ <h5>Our Sponsors</h5>
         <div class="container">
             <div class="row">
                 <div class="col-lg-12">
-                    <p>Copyright &copy; @messages("general.organization") 2016-2018</p>
+                    <p>Copyright &copy; @messages("general.organization") 2016-2019</p>
```
progwml6: are you able to make this be the current year so it doesn't need to be changed again?

felixoi: Fixed in latest commit

Katrix (approved):

PR #803, opened 2019-05-01
Fixed icon and updated dependencies

felixoi:


build.sbt
```diff
@@ -91,9 +91,10 @@ lazy val discourse = project.settings(
     "io.circe"                   %% "circe-core"           % circeVersion,
     "io.circe"                   %% "circe-generic-extras" % circeVersion,
     "io.circe"                   %% "circe-parser"         % circeVersion,
+    "com.typesafe.akka"          %% "akka-http"            % akkaHttpVersion,
```
Katrix, line 94: Why did you add this dep?

Katrix (commented):
Did you make sure all akka deps are the same?

build.sbt
felixoi: Because I had to force `akka-http-circe` to use this version of akka-http. Otherwise he get conflicting akka versions.

Katrix (approved):

PR #808, opened 2019-05-09
Resolve the icon url for a project statically

Katrix:
Currently when serving icons for projects on Ore, we set the content of the `src` attribute to the ore project icon endpoint. What does this endpoint do? 1. Fetches the project from the DB, 2. Checks if the project has an icon, 3. If it does, sends that, and if not, redirects to auth.

Why is this stupid? Because we have all the information we need already. So, a better approach implemented here, is to check if the project has an icon before generating the link, and only link to the icon endpoint if it does.

This should dramatically reduce the amount of requests needed to render the homepage (if no requests are present, it roughly halves it). It will also reduce the amount if times we hit the DB.


Katrix:
Superseded by #809

PR #822, opened 2019-05-17
Adds autoprefixer

felixoi:


Katrix:
Can we support ESR browsers too? Only know about Firefox's ESR branch. Are there others?

felixoi:
It's supporting latest Firefox ESR

Katrix:
Why doesn't it appear in the `autoPrefixerBrowsers` setting then?

Katrix (approved):

PR #828, opened 2019-05-25
Added warning for internet explorer

### LINKED ISSUE #378: Internet Exploder support?
Browser & Version: `IE (vomit) 11.125.16299.0 (11.0.49)`

image

Operating System: `Windoze`

image

Edge works: 

image

I just happened to type "internet explorer" out of habit after I needed a different browser to work around a bug, and couldn't be assed using incognito mode.

This bug really should be pending checking on the amount of hits the forum / home page get on Internet Explorer.

Checking Ore for browser stats would be counter productive however, as people don't tend to use broken shit if they can help it.

felixoi:
Fixes #378

### DIFF
--- a/ore/public/javascripts/main.js
+++ b/ore/public/javascripts/main.js
@@ -55,6 +55,10 @@ function toggleSpinner(e) {
 hljs.initHighlightingOnLoad();
 
 $(function() {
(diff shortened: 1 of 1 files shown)


Katrix (approved):

ryantheleach:
Quick question, how does the new IE/Edge that's based on chrome present itself on Ore?

felixoi:
Since it's based on chromium it will not get any warning and display everything correctly.

PR #836, opened 2019-06-13
Use vue for the homepage

### LINKED ISSUE #259: Total pages
There is only arrows for navigation. Please, add navigation using page numbers.
Like in github issues page:
<< 1 2 3 ... 99 >>

So, we can see, how much pages we have

felixoi:
Opened to keep track of the stuff. Not fully ready yet.

Fixes #259


build.sbt
```diff
@@ -190,8 +190,48 @@ def flexmarkDep(module: String) = {
...
+      "babel-loader"                       -> "8.0.6",
+      "@babel/core"                        -> "7.4.5",
+      "terser-webpack-plugin"              -> "1.3.0",
+      "mini-css-extract-plugin"            -> "0.7.0",
+      "optimize-css-assets-webpack-plugin" -> "5.0.1",
+      "sass-loader"                        -> "7.1.0",
+      "postcss-loader"                     -> "3.0.0",
+      "autoprefixer"                       -> "9.5.1",
+      "node-sass"                          -> "4.12.0",
+      "query-string"                       -> "6.5.0",
```
felixoi: This shouldnt be a dev dependency.

oreClient/src/main/resources/assets/components/ProjectList.vue
```diff
@@ -0,0 +1,223 @@
...
+            <i class="fas fa-spinner fa-spin"></i>
+            <span>Loading projects for you...</span>
+        </div>
+        <div v-show="!loading">
+            <ul class="list-group project-list">
+                <li v-for="project in projects" class="list-group-item project @entry.visibility.cssClass">
+                    <div class="container-fluid">
+                        <div class="row">
+                            <div class="col-xs-12 col-sm-1">
+                                <img class="user-avatar user-avatar-sm" :src="project.icon_url" :alt="project.name" />
```
felixoi: Note to myself to make this clickable again.

build.sbt
```diff
@@ -190,8 +190,48 @@ def flexmarkDep(module: String) = {
   "com.vladsch.flexmark" % artifactId % flexmarkVersion
 }
 
+lazy val oreClient = project
+  .enablePlugins(ScalaJSBundlerPlugin)
+  .settings(
+    name := "ore-client",
+    commonSettings,
+    useYarn := true,
+    scalacOptions += "-P:scalajs:sjsDefinedByDefault",
```
Katrix: Can probably delete this

ore/public/javascripts/apiRequests.js
```diff
@@ -3,31 +3,23 @@
...
 let isLoggedIn = false;
 
+//=====> SETUP
+
+jQuery.ajaxSettings.traditional = true;
+
 //=====> HELPER FUNCTIONS
 
-function apiV2Request(url, method, data, isRetry) {
+function apiV2Request(url, method = "GET", data = {}, isRetry = true) {
```
Katrix: From what I remember when I wrote this function, isRetry should only be true when it has already tried the request once, failed, and gotten a new token. As it is now it breaks loading if the user isn't logged in.

oreClient/src/main/resources/assets/Home.vue
```diff
@@ -0,0 +1,205 @@
+<template>
+    <div class="row">
+        <div class="col-md-9">
+            <div class="project-search">
+                <input type="text" class="form-control" v-model="q" :placeholder="'Search in ' + projectCountPlaceholder + ' projects, proudly made by the community...'" />
```
Katrix: Do we want to update the project count when a filter is applied too?

oreClient/src/main/resources/assets/components/Pagination.vue
```diff
@@ -0,0 +1,84 @@
+<template>
```
Katrix, line 1: Can we make a prettier pagination for the home page? It just doesn't feel like it fits at all. The colors are all wrong, and it feels so small. At least make it a bit bigger/longer.

Katrix, line 1: This breaks when on the second page and searching for something which only returns one page.

ore/public/javascripts/apiRequests.js
Katrix: From what I remember when I wrote this function, isRetry should only be true when it has already tried the request once, failed, and gotten a new token. As it is now it breaks loading if the user isn't logged in.

oreClient/src/main/resources/assets/components/Pagination.vue
felixoi: Yeah just copied the design for the version page

oreClient/src/main/resources/assets/Home.vue
felixoi: Isn't this updating? It's bound to the pagination.count of the last request. So if a filter is applied pagination.count should have a different value or not?

ore/public/javascripts/apiRequests.js
felixoi: Yeah indeed. Default should be false.

Katrix: Actually, when I think about it a bit more, the whole idea of "Fire off a request and see if it succeeds" is stupid when the authentication request includes the expiration date. Maybe rewrite stuff to prefer to use that?

oreClient/src/main/resources/assets/Home.vue
Katrix: What I more meant is if this is the behavior we want, or if we want the count to  only change when no filters are used.

felixoi: I'd say we should update it. Will push a commit later I've locally already which changes the message if filters are active but still updates the count.

felixoi: Can you recheck if this fits now?

Katrix (approved):

PR #840, opened 2019-06-18
Use correct format for auth header

Katrix:
Fixes the issue Phit pointed out about Ore not respecting the spec in regards to the Authorization header.

It's now `Authorization: OreApi session=foo` and `Authorization: OreApi apikey=bar`


phit:
I believe the format should be `Authorization: OreApi session="foo"`. Also what happens if you do `Authorization: OreApi apikey="bar", session="foo"`

Katrix:
Think the quotes are optional. As for what happens if you specify both, it will grab the one it needs, and ignore the other. We don't handle parsing the header ourselves.

felixoi (approved):

PR #849, opened 2019-07-04
Fixed evolutions

felixoi:


ore/app/OreApplicationLoader.scala
```diff
@@ -65,12 +66,14 @@ class OreComponents(context: ApplicationLoader.Context)
...
     with SlickComponents
-    with SlickEvolutionsComponents {
+    with SlickEvolutionsComponents
+    with EvolutionsComponents {
   val prefix                                = "/"
   override lazy val router: Router          = wire[_root_.router.Routes]
   lazy val apiV2Routes: _root_.apiv2.Routes = wire[_root_.apiv2.Routes]
 
   use(prefix) //Gets around unused warning
+  applicationEvolutions
```
Katrix: Can you write something like `eager(applicationEvolutions)`? Same semantics, but makes it more clear that it does stuff at startup.

version.txt
```diff
@@ -0,0 +1,6 @@
+[0m[[0m[0minfo[0m] [0m[0mLoading settings for project global-plugins from idea.sbt,sbt-updates.sbt ...[0m
```
Katrix: Can you remove this?

PR #855, opened 2019-07-23
Only ship the font-awesome icons we use

felixoi:


orePlayCommon/app/views/layout/base.scala.html
```diff
@@ -37,7 +37,7 @@
 
         @if(scriptsEnabled) {
             <script type="text/javascript" src="@routes.Assets.versioned("lib/jquery/dist/jquery.min.js")"></script>
-            <script defer type="text/javascript" src="@routes.Assets.versioned("lib/font-awesome/js/all.min.js")" data-auto-add-css="false"></script>
+            <script type="text/javascript" src="@routes.Assets.versioned("vue/font-awesome.js")"></script>
```
Katrix: Why not defer?

felixoi: Actually we could load it with defer but the problem is that the page is then loaded faster then the script. Looks kinda stupid in the first moment after the page load then.

PR #856, opened 2019-07-25
Mention yarn requirement in readme

felixoi:


Katrix:
Can you also mention the need for Node, and the full path of the config folder?

Katrix:
One note that we should mention is that the application.conf file needs to be refreshed now and then as we add config stuff not too rarely

Grinch (approved):
Changes look good!

PR #865, opened 2019-08-01
Use linesIterator instead of lines

Katrix:
Should fix compiling with JDK 11


Grinch (approved):
Appears to compile properly with JDK 8 and JDK 11.

PR #871, opened 2019-08-14
Run assets on travis

Katrix:
Runs the assets pipeline of travis.

Also checks the formatting globally, so that we don't forget to add stuff to it later.

PR #872, opened 2019-08-17
Build scss on oreClient

felixoi:
also includes https://github.com/SpongePowered/Ore/pull/871 to test travis


Katrix (approved):
Broke goto definition in the templates for CSS stuff, but otherwise good. Let's do this then, and hope we don't need it for too long.

PR #904, opened 2019-09-14
Added matomo

felixoi:


simon816:
We removed Google Analytics last year, see https://github.com/SpongePowered/SpongeHome/issues/147
Has anything changed since then?

felixoi:
@simon816 yes.
1. We self host the matomo instance, so no data leaves our servers
2. We setup IP's and stuff to be anonymized

Also currently we only track where the user was coming from and the user count.

Issue #958, opened 2020-03-16
Category icons of resources are not updated on page change


AuroraLS3:
**Describe the bug**

This is a minor visual bug.

The resource category icons are not updated when page is changed. (Note that the `title` property changes appropriately)

Page 1:
![Page 1](https://user-images.githubusercontent.com/24460436/76760860-1cda1b80-6797-11ea-8677-fad8bdd41fa2.png)
Page 2:
![Page 2](https://user-images.githubusercontent.com/24460436/76760884-2c596480-6797-11ea-9554-fc824ff1e146.png)

**To Reproduce**
1. Go to https://ore.spongepowered.org/
2. Observe category icon of a resource
3. Click on Pagination to move to another page (say '2' at the bottom)
4. Observe that the category icon of the resource did not change
5. To confirm: Refresh the page
6. Observe that the proper category icon was loaded

**Expected behavior**
The category icon should also change when the resource is loaded

**Environment **
- OS: Windows 10
- Browser: Firefox 74.0

**Additional context**
Staging server could not be accessed to check if the bug was already fixed. (`SSL_ERROR_NO_CYPHER_OVERLAP`)

Katrix:
Closed by #960

PR #960, opened 2020-03-26
Added vue-fontawesome

felixoi:


Katrix:
Added it in other places too. Also did a few other fixes. Fixes #958

PR #990, opened 2020-04-26
Project Creation Vue

felixoi:


oreClient/src/main/assets/stores/top.js
```diff
@@ -13,6 +13,11 @@ const mutations = {
             message: payload.message
         })
     },
+    addAlerts(state, payload) {
+        state.alerts[payload.level] = state.alerts[payload.level].concat(payload.messages.map(m => {
```
Katrix, line 19: Is this safe in terms of reactivity and Vue?

oreClient/src/main/assets/pages/project/NewProject.vue
```diff
@@ -0,0 +1,144 @@
...
+                        </div>
+
+                        <div class="form-group">
+                            <label for="projectPluginId">Plugin id</label>
+                            <input type="text" id="projectPluginId" name="pluginId" class="form-control" v-model.trim="pluginId" required>
+                        </div>
+
+                        <div class="form-group">
+                            <label for="projectCategory">Project category</label>
+                            <select class="form-control" v-model="category" required>
```
Katrix: Missing id

oreClient/src/main/assets/api.js
```diff
@@ -27,6 +28,11 @@ export class API {
                         withCredentials: true
                     },
                 }).done((data) => {
+                    //store.commit({
```
Katrix: Did you ever get this working?

oreClient/src/main/assets/pages/project/NewProject.vue
```diff
@@ -0,0 +1,144 @@
...
+                    });
+                } else {
+                    API.request("projects", "POST", {
+                        name: this.projectName,
+                        plugin_id: this.pluginId,
+                        category: this.category,
+                        description: this.projectDescription,
+                        owner_name: this.owner
+                    }).then((data) => {
+                        this.$router.push({path: `/${data.namespace.owner}/${data.namespace.slug}`})
```
Katrix: Why not do name based routing, and also pass it extra info we have about the project. We have the whole fetched project, so might as well just pass it in. Lookup by namespace is also a bit slower currently.

oreClient/src/main/assets/api.js
felixoi: yeah that works, remove the comments and it should work instantly

PR #1032, opened 2020-11-24
Scope changes and other misc fixes

Katrix:
Keep scope info for a bit longer. Should save us a few joins here and there. Won't mean too much, but improves the code quality a bit.


Katrix:
Reworked and updated in #1062

PR #1035, opened 2020-11-28
Some more misc fixes

Katrix:
Some more fixes I did while recording the demo


Katrix:
Stupid node-sass failing the build. Will deal with that later

PR #1038, opened 2021-02-23
Minor fixes

LoboMetalurgico:


Katrix:
We'd love to take in your changes into upstream here. However, before we do so, there are a few things we ask you to fix.
1. Keep PowerNukkit specific stuff in your repo
2. Keep the default settings config files. The intention is that for repo specific stuff, you're supposed to override the changes in `application.conf.template`. For user specific changes, that's what the user puts in their copy of `application.conf` after copying and renaming `application.conf.template`.

LoboMetalurgico:
Sorry, this change was for my fork, I accidentally threw it here. Sorry.

PR #1039, opened 2021-06-23
add support for API8 sponge plugins

Faithcaio:
We need API8 plugin support sooner than later.


Katrix:
Please use Circe for json parsing if we're going to just patch this into the current Ore version

PR #1043, opened 2021-09-04
Update for plugin-meta 0.8.0 for API 8 meta changes

dualspiral:
This is to support the upcoming changes in Sponge API 8+

* Updated plugin-meta to 0.8.0-SNAPSHOT 
* Added plugin-meta:mcmod-info for legacy plugin/mod handling
* Cleaned up the Json file handler to use plugin-meta
* Updated other areas that used plugin meta (specifically for the removal of the `get` prefix)

Mostly want to make sure I've done the right thing here, mostly with formatting and idioms etc. Shouldn't be merged just yet, I'd rather release meta 0.8 proper and have Ore use that, but I don't imagine there will be any more changes.

### DIFF
--- a/orePlayCommon/app/ore/models/project/io/PluginFileData.scala
+++ b/orePlayCommon/app/ore/models/project/io/PluginFileData.scala
@@ -1,19 +1,18 @@
 package ore.models.project.io
 
 import scala.language.higherKinds
-
 import java.io.BufferedReader
-import com.google.gson.stream.JsonReader
 
 import scala.collection.mutable.ArrayBuffer
 import scala.jdk.CollectionConverters._
+import scala.jdk.OptionConverters._
 import scala.util.control.NonFatal
-
 import ore.data.project.Dependency
 import ore.db.{DbRef, Model, ModelService}
 import ore.models.project.{TagColor, Version, VersionTag}
-
 import org.spongepowered.plugin.meta.McModInfo
+import org.spongepowered.plugin.metadata.builtin.MetadataParser
+import org.spongepowered.plugin.metadata.model.PluginDependency
 
 /**
   * The metadata within a [[PluginFile]]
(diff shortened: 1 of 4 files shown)


dualspiral:
I spoke too soon, there may be more changes

image

orePlayCommon/app/ore/models/project/io/PluginFileData.scala
```diff
@@ -211,96 +210,34 @@ object ModTomlHandler extends FileTypeHandler("mod.toml") {
...
-      dataValues.toSeq
+      val metadata    = MetadataParser.read(bufferedReader)
+      val firstPlugin = metadata.metadata().asScala.head
+      dvs += StringDataValue("id", firstPlugin.id)
+      dvs += StringDataValue("version", firstPlugin.version.toString)
+      firstPlugin.name.toScala.map(dvs += StringDataValue("name", _))
+      firstPlugin.description.toScala.map(dvs += StringDataValue("description", _))
+      dvs += StringListValue("authors", firstPlugin.contributors().asScala.map(_.name()).toSeq)
+      dvs += DependencyDataValue("dependencies", readDependencies(firstPlugin.dependencies().asScala))
+      dvs.toSeq
```
Katrix: Something like this would be preferable
```scala
Seq[DataValue](
  StringDataValue("id", firstPlugin.id),
  StringDataValue("version", firstPlugin.version.toString),
  StringListValue("authors", firstPlugin.contributors().asScala.map(_.name()).toSeq),
  DependencyDataValue("dependencies", readDependencies(firstPlugin.dependencies().asScala))
) ++ Seq(
  firstPlugin.name.toScala.map(v => StringDataValue("name", v)),
  firstPlugin.description.toScala.map(v => StringDataValue("description", v))
).flatten
```

orePlayCommon/app/ore/models/project/io/PluginFileData.scala
```diff
@@ -211,96 +210,34 @@ object ModTomlHandler extends FileTypeHandler("mod.toml") {
...
     } catch {
       case NonFatal(e) =>
         e.printStackTrace()
         Nil
     }
   }
+
+  def readDependencies(in: Iterable[PluginDependency]): Seq[Dependency] = {
+    in.map(dep =>
```
Katrix: Braces for multi line lamda

orePlayCommon/app/ore/models/project/io/PluginFileData.scala
```diff
@@ -211,96 +210,34 @@ object ModTomlHandler extends FileTypeHandler("mod.toml") {
...
+        Dependency(dep.id, if (dep.version.hasRestrictions) {
```
Katrix: `Option.when` instead of if-else

Katrix (changes requested):

PR #1062, opened 2022-06-01
Scope changes and other misc fixes 

Katrix:
Keep scope info for a bit longer. Should save us a few joins here and there. Won't mean too much, but improves the code quality a bit.

Rebased and cleaned up version of #1032


Katrix:
Merged manually

PR #1064, opened 2022-06-01
Don't use exact query when getting projects by namespace

Katrix:
IIRC exact is a relic from when the API worked on plugin ids, not project namespaces, when looking up projects. We don't need to use exact here anymore


Katrix:
Merged manually

